PR #13, opened 2023-04-27
Implement a SpeechToText Provider

julien-nc:
Here is an STT provider using OpenAI's Whisper API. It supports mp3, mp4, mpeg, mpga, m4a, wav, or webm files as mentioned [there](https://platform.openai.com/docs/api-reference/audio/create).

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -13,6 +13,7 @@ use OCA\OpenAi\Listener\OpenAiReferenceListener;
 use OCA\OpenAi\Reference\ChatGptReferenceProvider;
 use OCA\OpenAi\Reference\ImageReferenceProvider;
 use OCA\OpenAi\Reference\WhisperReferenceProvider;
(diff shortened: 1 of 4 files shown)


lib/AppInfo/Application.php
```diff
@@ -48,6 +49,10 @@ public function register(IRegistrationContext $context): void {
 			$context->registerReferenceProvider(ImageReferenceProvider::class);
 			$context->registerReferenceProvider(WhisperReferenceProvider::class);
 			$context->registerEventListener(RenderReferenceEvent::class, OpenAiReferenceListener::class);
+
+			if (version_compare($this->config->getSystemValueString('version', '0.0.0'), '27.0.0', '>=')) {
```
marcelklehr, line 53: Won't this, pre-v27, still load the STTProvider php file and then fail at the usage of the non-existing interface?

julien-nc: I thought so too and tried and it works. I guess the STTProvider.php file is only loaded if it's actually used at run time.

PR #52, opened 2023-11-02
AUTH: Support basic auth for local ai

mickenordin:
By default localai.io does not come with any authentication.

It is however trival to run behind a reverse proxy, and the most common way to authenticate with any webserver is with the widely supported basic auth scheme, using a username and password.

This patch allows an administrator or user to set a username and password to use with basic auth for a local installation of localai.


mickenordin:
@julien-nc I was hoping you would like to review this one? 🙏

MB-Finski:
Thanks for the contribution! ... and sorry for not reviewing this immediately; I'll get on it now!

MB-Finski:
Hey, @mickenordin: Thanks again for the contribution! If I'm not mistaken the personal settings front-end changes got dropped when merging the changes from main. Could, please, you add those back so that the attribution will be correct?

Storing the setting broke with 1.1.* due to the separate settings class, but I already made the necessary changes while testing this branch so no need to bother with those; I'll merge the changes to main once we get this merged (or I can create a pr on your branch if you'd like).

Another random though was that if  the api key is defined the basic authentication is skipped silently which can cause some confusion. Maybe in case of a 401 when both are defined we should add a warning message about this, WDYT?

mickenordin:
> Hey, @mickenordin: Thanks again for the contribution! If I'm not mistaken the personal settings front-end changes got dropped when merging the changes from main. Could you, please, add those back so that the attribution will be correct?
> 
Oops, sorry about that...
> Storing the settings broke with 1.1.* due to the separate settings class I introduced, but I already made the necessary changes while testing this branch so no need to bother with those; I'll merge the changes to main once we get this merged (or I can create a pr on your branch if you'd like).
> 
I gave you write access on my fork so you can do what ever you need to :)

> Another random thought was that if the api key is defined the basic authentication is skipped silently which can cause some confusion. Maybe in case of a 401 when both are defined we should add a warning message about this, WDYT?

Yes, I am not sure on how to  best handle this, I thought it made sense to give the api key precedence, but maybe we can do something more clever here, like make the options mutually exclusive in the ui. I am not very skilled in vue, but I am sure it would be possible to work something out?

MB-Finski:
Hmm.. It gets a bit tricky when making the setting mutually exclusive since the personal settings allow also filling both methods and in that case the admin api key would override the user specified basic auth and vice versa. I think the reasonable thing to do here is to have the admin manually specify which auth method has to be used and then that option is enforced in personal settings, too.

mickenordin:
> @mickenordin What do you think about the UI changes? Now the admin can choose the authentication method and then this auth method is forced in the personal settings, too. All these extra settings are hidden if openAI is used.

I will build a new version, and try it out!

mickenordin:
Yes, that looks perfect! Nice job @MB-Finski !

MB-Finski (approved):

PR #54, opened 2023-11-03
Add support for quota limiting, text gen token limiting and prompt history clearing

MB-Finski:
Add support for quota limiting (text, images and translation), expose text gen new token limit as an admin setting and allow users to clear prompt history from personal settings.

Quota limit is instance wide from admin settings but can be by passed by individual users by providing an own API key in the personal settings. Users can view their current quota use in personal settings.

Local storage of images was dropped in favor of implementing that in the up-coming unified text2image_helper.


appinfo/info.xml
```diff
@@ -71,14 +71,14 @@ Negative:
...
-* the trained model is freely available, and thus can be ran on-premises
+* the trained model is freely available, and thus can be ran on-premises 
 
 Negative:
 * the training data is not freely available, limiting the ability of external parties to check and correct for bias or optimise the model’s performance and CO2 usage.
 
 Learn more about the Nextcloud Ethical AI Rating [in our blog](https://nextcloud.com/blog/nextcloud-ethical-ai-rating/).
 ]]></description>
-    <version>1.0.13</version>
+    <version>1.0.14</version>
```
marcelklehr, line 81: If you add features, you should bump the minor version, so it'd be `1.1.0` (see https://semver.org)

lib/Db/QuotaUsage.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->addType('type', 'integer');
+		$this->addType('units', 'integer');
+		$this->addType('timestamp', 'integer');
+	}
+
+	#[\ReturnTypeWillChange]
+	public function jsonSerialize() {
+		return [
+			'id' => $this->id,
+            'user_id' => $this->userId,
```
marcelklehr, line 62: Please run `composer run cs.fix` and commit the result to fix this kind of coding style issues

lib/Db/QuotaUsageMapper.php
```diff
@@ -0,0 +1,302 @@
...
+				'type' => $qb->createNamedParameter($type, IQueryBuilder::PARAM_INT),
+				'user_id' => $qb->createNamedParameter($userId, IQueryBuilder::PARAM_STR),
+				'units' => $qb->createNamedParameter($units, IQueryBuilder::PARAM_INT),
+				'timestamp' => $qb->createNamedParameter($timeStamp, IQueryBuilder::PARAM_INT),
+			]);
+
+		$qb->executeStatement();
+		$qb->resetQueryParts();
+
+		return $this->getQuotaUsageOfUser((int)$qb->getLastInsertId(), $userId);
```
marcelklehr, line 227: This seems to duplicate QBMapper#insert(Entity)

lib/Migration/Version010014Date20231008173034.php
```diff
@@ -0,0 +1,91 @@
...
+				'notnull' => true,
+			]);
+			$table->addColumn('units', Types::INTEGER, [
+				'notnull' => true,
+			]);			
+			$table->addColumn('timestamp', Types::INTEGER, [
+				'notnull' => true,
+			]);
+			$table->setPrimaryKey(['id']);
+			$table->addIndex(['user_id'], 'openai_quota_usage_userid');
```
marcelklehr, line 78: type and timestamp should probably have indeces, too, and also in combination, based on how you query the table in the mapper.

lib/Service/OpenAiAPIService.php
```diff
@@ -85,6 +87,114 @@ public function getPromptHistory(string $userId, int $type): array {
...
+			//  Unlimited quota:
+			return false;
+		}
+
+		$quotaPeriod = intval($this->config->getAppValue(Application::APP_ID, 'quota_period', Application::DEFAULT_QUOTA_PERIOD));
+		
+		try {
+			$quotaUsage= $this->quotaUsageMapper->getQuotaUnitsOfUserInTimePeriod($userId, $type, $quotaPeriod);
+		} catch (DoesNotExistException | MultipleObjectsReturnedException | Exception | RuntimeException $e) {
+			throw new Exception('Could not retrieve quota usage for user: ' . $userId . ' and quota type: ' . $type . '. Error: ' . $e->getMessage());
```
marcelklehr, line 149: You can pass the exception that triggered this catch to the new exception as a param to get a chained stack trace.

lib/Service/OpenAiAPIService.php
```diff
@@ -85,6 +87,114 @@ public function getPromptHistory(string $userId, int $type): array {
...
+
+	/**
+	 * @param string $userId
+	 * @return array
+	 */
+	public function getUserQuotaInfo(string $userId): array {
+		// Get quota limits (if the user has specified an own OpenAI API key, no quota limit, just supply default values as fillers)
+		$quotas = $this->hasOwnOpenAiApiKey($userId) ?
+				Application::DEFAULT_QUOTAS :
+				json_decode($this->config->getAppValue(Application::APP_ID, 'quotas', json_encode(Application::DEFAULT_QUOTAS)));
```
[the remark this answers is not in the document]
marcelklehr: Also, note that json_decode uses StdClass by default instead of an associative array so $quotas will now have the type `\StdClass[]|array[]` which is kind of unwieldy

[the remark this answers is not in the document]
marcelklehr: This is a good reason why not using an AI coding assistant and looking up the function's docs instead is often a good idea ;)

lib/Service/OpenAiAPIService.php
```diff
@@ -85,6 +87,114 @@ public function getPromptHistory(string $userId, int $type): array {
...
+		// Get quota period
+		$quotaPeriod = intval($this->config->getAppValue(Application::APP_ID, 'quota_period', Application::DEFAULT_QUOTA_PERIOD));
+		// Get quota usage for each quota type:
+		$quotaInfo = [];
+		foreach (Application::QUOTA_TYPES as $quotaType => $quotaTypeName) {
+			$quotaInfo[$quotaType]['type'] = $quotaTypeName;
+			$quotaInfo[$quotaType]['used'] = $this->quotaUsageMapper->getQuotaUnitsOfUserInTimePeriod($userId, $quotaType, $quotaPeriod);
+			$quotaInfo[$quotaType]['limit'] = intval($quotas[$quotaType]->value);
```
marcelklehr, line 171: this will fail sometimes, see above

lib/AppInfo/Application.php
```diff
@@ -39,6 +39,33 @@ class Application extends App implements IBootstrap {
...
+	public const DEFAULT_MAX_NUM_OF_TOKENS = 1000;
+	public const DEFAULT_QUOTA_PERIOD = 30;
+	
+	public const QUOTA_TYPE_TEXT = 0;
+	public const QUOTA_TYPE_IMAGE = 1;
+	public const QUOTA_TYPE_TRANSCRIPTION = 2;
+
+	
+	public const QUOTA_TYPES = [
+		self::QUOTA_TYPE_TEXT => 'text',
```
marcelklehr, line 51: This is a bit annoying, I feel, YMMV but I feel like simply always using a string as an identifier would simplify the logic a lot.

lib/AppInfo/Application.php
```diff
@@ -39,6 +39,33 @@ class Application extends App implements IBootstrap {
...
+		self::QUOTA_TYPE_IMAGE => 'image',
+		self::QUOTA_TYPE_TRANSCRIPTION => 'transcription',
+	];
+	public const DEFAULT_QUOTAS = [
+		self::QUOTA_TYPE_TEXT => ['type' => self::QUOTA_TYPES[self::QUOTA_TYPE_TEXT], 'value' => 0], // 0 = unlimited
```
marcelklehr, line 56: Is there a reason that the value of the array key has a type key with the value of the array key again? Couldn't we do 

```suggestion
		self::QUOTA_TYPE_TEXT =>  0, // 0 = unlimited
```

instead?

lib/Service/OpenAiAPIService.php
```diff
@@ -121,22 +249,39 @@ public function createCompletion(?string $userId, string $prompt, int $n, string
 	 * @param bool $storePrompt
 	 * @return array|string[]
 	 * @throws \OCP\DB\Exception
+	 * @throws Exception
 	 */
 	public function createChatCompletion(?string $userId, string $prompt, int $n, string $model, int $maxTokens = 1000,
 										 bool $storePrompt = true): array {
+		if($this->isQuotaExceeded($userId, Application::QUOTA_TYPE_TEXT)){
+			return ['error' => 'Text generation quota exceeded'];
```
marcelklehr, line 257: Aaah

lib/Service/OpenAiAPIService.php
```diff
@@ -182,15 +327,27 @@ public function transcribeFile(?string $userId, File $file, bool $translate = fa
 	 */
 	public function transcribe(?string $userId, string $audioFileContent, bool $translate = true,
 							   string $model = Application::DEFAULT_TRANSCRIPTION_MODEL_ID): array {
+		if($this->isQuotaExceeded($userId, Application::QUOTA_TYPE_TRANSCRIPTION)){
+			return ['error' => 'Transcription usage quota exceeded'];
```
marcelklehr, line 331: Aaah

lib/Service/OpenAiAPIService.php
```diff
@@ -202,6 +359,9 @@ public function transcribe(?string $userId, string $audioFileContent, bool $tran
 	 * @throws \OCP\DB\Exception
 	 */
 	public function createImage(?string $userId, string $prompt, int $n = 1, string $size = Application::DEFAULT_IMAGE_SIZE): array {
+		if($this->isQuotaExceeded($userId, Application::QUOTA_TYPE_IMAGE)){
+			return ['error' => $this->l10n->t('Image generation quota exceeded')];
```
marcelklehr, line 363: Aaaah

lib/Settings/Admin.php
```diff
@@ -29,18 +29,36 @@ public function getForm(): TemplateResponse {
 		$textPickerEnabled = $this->config->getAppValue(Application::APP_ID, 'text_completion_picker_enabled', '1') === '1';
 		$translationProviderEnabled = $this->config->getAppValue(Application::APP_ID, 'translation_provider_enabled', '1') === '1';
 		$sttProviderEnabled = $this->config->getAppValue(Application::APP_ID, 'stt_provider_enabled', '1') === '1';
+		
+		$maxGeneratedTokens = $this->config->getAppValue(Application::APP_ID, 'max_tokens', Application::DEFAULT_MAX_NUM_OF_TOKENS) ?: Application::DEFAULT_MAX_NUM_OF_TOKENS;
+		$quotaPeriod = $this->config->getAppValue(Application::APP_ID, 'quota_period', Application::DEFAULT_QUOTA_PERIOD) ?: Application::DEFAULT_QUOTA_PERIOD;
+		$quotas = json_decode($this->config->getAppValue(Application::APP_ID, 'quotas', json_encode(Application::DEFAULT_QUOTAS)) ?: json_encode(Application::DEFAULT_QUOTAS));
```
marcelklehr, line 35: A settings service would be nice for this as well

lib/Settings/Admin.php
```diff
@@ -29,18 +29,36 @@ public function getForm(): TemplateResponse {
...
+
+		// Make sure all quota types are set in the json encoded app value
+		if(count($quotas) !== count(Application::QUOTA_TYPES)) {
+			foreach(Application::QUOTA_TYPES as $type => $typeName) {
+				if(!isset($quotas[$type])) {
+					$quotas[$type] = Application::DEFAULT_QUOTAS[$type];
+				}
+			}
+			$this->config->setAppValue(Application::APP_ID, 'quotas', json_encode($quotas));
+		}
```
marcelklehr, line 45: This should be in the settings service, too

lib/TextProcessing/HeadlineProvider.php
```diff
@@ -16,6 +16,7 @@ public function __construct(
 		private OpenAiAPIService $openAiAPIService,
 		private IConfig $config,
 		private IL10N $l10n,
+		private ?string $userId,
```
marcelklehr, line 19: Doesn't work yet, because text processing providers potentially run in a background job which doesn't have a session

lib/TextProcessing/ReformulateProvider.php
```diff
@@ -15,6 +15,7 @@ public function __construct(
 		private OpenAiAPIService $openAiAPIService,
 		private IConfig $config,
 		private IL10N $l10n,
+		private ?string $userId,
```
marcelklehr, line 18: Doesn't work yet, see above.

src/components/AdminSettings.vue
```diff
@@ -88,9 +89,77 @@
...
+									{{ t('integration_openai', capitalizedWord(quota.type)) }}
+								</td>
+								<td>
+									<input :id="'openai-api-quota-' + quota.type"
+										v-model="quota.value"
+										:title="t('integration_openai', 'A per-user limit for uasge of this API type (0 for unlimited))')"
+										type="number"
+										@input="onInput(false)">
+									<span v-if="quotaInfo !== null" class="text-cell">
+										{{ t('integration_openai', quotaInfo[index].unit) }}
```
marcelklehr, line 135: This doesn't work. To be translatable, the call to t must be with a string literal as a message

src/components/PersonalSettings.vue
```diff
@@ -16,31 +16,84 @@
...
+						</tr>
+					</thead>
+					<tbody>
+						<tr v-for="quota in quotaInfo.quota_usage" :key="quota.type">
+							<td>{{ t('integration_openai', capitalizedWord(quota.type)) }} </td>
+							<td v-if="quota.limit > 0">
+								{{ quota.used / quota.limit * 100 + ' %' }}
+							</td>
+							<td v-else>
+								{{ t('integration_openai',quota.used + ' ' + quota.unit) }}
```
marcelklehr, line 77: Doesn't work, see docs on using the translation functions with paramters

marcelklehr:
Do you have screenshots of the front-end changes, before and after?

marcelklehr:
For reviewing it's often easier to do one feature per pull request, rather than multiple ones

lib/Db/QuotaUsageMapper.php
MB-Finski: Hmm. True! This was my first dabble with having something stored in the db so I'll blame that.

lib/Db/PromptMapper.php
```diff
@@ -182,6 +182,26 @@ public function deleteUserPrompts(string $userId): void {
...
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete($this->getTableName())
+			->where(
+				$qb->expr()->eq('user_id', $qb->createNamedParameter($userId, IQueryBuilder::PARAM_STR))
+			)
+			->andWhere(
+				$qb->expr()->eq('type', $qb->createNamedParameter($type, IQueryBuilder::PARAM_INT))
+			);
+		$qb->executeStatement();
+		$qb->resetQueryParts();
```
marcelklehr, line 202: not necessary, $qb will be out of scope after this statement

lib/Controller/OpenAiAPIController.php
```diff
@@ -157,4 +160,53 @@ public function getImageGenerationPage(string $hash): TemplateResponse {
...
+	 * @return DataResponse
+	 */
+	#[NoAdminRequired]
+	public function clearPromptHistory(?bool $clearTextPrompts = null, ?bool $clearImagePrompts = null): DataResponse {
+		$this->logger->debug('Clearing prompt history: ' . 'clearTextPrompts ' . strval($clearTextPrompts) . ' clearImagePrompts ' . strval($clearImagePrompts));
+		if ($clearTextPrompts === True) {			
+			try {
+				$this->openAiAPIService->clearPromptHistory($this->userId, Application::PROMPT_TYPE_TEXT);
+			} catch (Exception $e) {				
+				return new DataResponse(['error' => $e->getMessage()],Http::STATUS_BAD_REQUEST);
```
marcelklehr, line 176: I'd say a Database exception is rather a 500 than a 400, no?

lib/Service/OpenAiAPIService.php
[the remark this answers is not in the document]
MB-Finski: Huh, that's an interesting default behavior, btw. To keep things in line with js syntax?

lib/AppInfo/Application.php
MB-Finski: Type is not actually equal to the array key, which is numeric but the text representation.  Here I semi-transitioned to using associative arrays which is kinda funny as I didn't go all the way. The point was to have a text representation on the js side. In my C oriented mind I was avoiding having a separate enum (that would need to match the "enum" here) on the js side so as to keep the code understandable.

lib/AppInfo/Application.php
MB-Finski: I think why I did this is still carry over from C. I'm still not fully used to dealing with associative arrays and fall back on my old habits as this would have been the way to do it without associative arrays.

lib/AppInfo/Application.php
MB-Finski: Fixed in https://github.com/nextcloud/integration_openai/pull/54/commits/44720e0aebfd3f549a6d9bd48a78df2872f26110

lib/AppInfo/Application.php
MB-Finski: Now types are numbers all the way and there's a separate translate method for the human readable unit/type in OpenAiApiService: https://github.com/nextcloud/integration_openai/pull/54/commits/44720e0aebfd3f549a6d9bd48a78df2872f26110

lib/Db/QuotaUsageMapper.php
MB-Finski: Thanks! Fixed in https://github.com/nextcloud/integration_openai/pull/54/commits/aa2052b27800e562f337992f530867d0f1e1b580

lib/Service/OpenAiAPIService.php
MB-Finski: Thanks! Fixed in: https://github.com/nextcloud/integration_openai/pull/54/commits/44720e0aebfd3f549a6d9bd48a78df2872f26110

lib/Service/OpenAiAPIService.php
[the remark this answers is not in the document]
MB-Finski: Thanks! Fixed in https://github.com/nextcloud/integration_openai/pull/54/commits/44720e0aebfd3f549a6d9bd48a78df2872f26110

lib/Migration/Version010014Date20231008173034.php
MB-Finski: Thanks! Fixed in https://github.com/nextcloud/integration_openai/pull/54/commits/8f6911e43a125fca257d1321194377723b17265b

lib/Db/PromptMapper.php
MB-Finski: Thanks! Fixed.

lib/Service/OpenAiAPIService.php
MB-Finski: Thanks, done!

lib/Service/OpenAiAPIService.php
```diff
@@ -94,23 +204,41 @@ public function getPromptHistory(string $userId, int $type): array {
...
-			'max_tokens' => $maxTokens,
-			'n' => $n,
-		];
-		if ($userId !== null) {
-			$params['user'] = $userId;
-		}
-		if ($storePrompt) {
+		
+	if($this->isQuotaExceeded($userId, Application::QUOTA_TYPE_TEXT)){
+		return ['error' => 'Text generation quota exceeded'];
```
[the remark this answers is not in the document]
MB-Finski: So, this method of passing errors was used also elsewhere in the existing code and that's why I basically copied it, as it also matched the front-end. I rewrote the error handling logic for the integration to have everything just throw base exception with a client friendly error message that is then passed on in the controller. Due to these error messages being passed on to the client, I hid error messages from the db mappers, for example.

Fixed in: https://github.com/nextcloud/integration_openai/pull/54/commits/6697be3b38618938a3fe9e5ab264d1eb7aeec0fe

lib/Service/OpenAiAPIService.php
MB-Finski: Fixed in: https://github.com/nextcloud/integration_openai/commit/6697be3b38618938a3fe9e5ab264d1eb7aeec0fe

lib/Service/OpenAiAPIService.php
MB-Finski: Fixed in: https://github.com/nextcloud/integration_openai/commit/6697be3b38618938a3fe9e5ab264d1eb7aeec0fe

lib/Service/OpenAiAPIService.php
MB-Finski: Fixed in: https://github.com/nextcloud/integration_openai/commit/6697be3b38618938a3fe9e5ab264d1eb7aeec0fe

lib/Service/OpenAiAPIService.php
```diff
@@ -182,15 +327,27 @@ public function transcribeFile(?string $userId, File $file, bool $translate = fa
...
-		return $this->request($userId, $endpoint, $params, 'POST', $contentType);
+
+		$response = $this->request($userId, $endpoint, $params, 'POST', $contentType);
+		if (isset($response['segments'])) {
+			$audioDuration = intval(round(floatval(array_pop($response['segments'])['end'])));
+			// Duration no longer needed:
+			unset($response['segments']);
+			$this->quotaUsageMapper->createQuotaUsage($userId, Application::QUOTA_TYPE_TRANSCRIPTION, $audioDuration);
+		}
+		return $response;
```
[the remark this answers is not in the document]
MB-Finski: Fixed in: https://github.com/nextcloud/integration_openai/commit/6697be3b38618938a3fe9e5ab264d1eb7aeec0fe

lib/Controller/OpenAiAPIController.php
MB-Finski: I changed the error handling quite a bit; do you think it would be worthwhile to add the proper code to each exception? I get that WAFs can filter traffic based on the response code but 400 should be quite innocuous in that sense...? Other than that the consumers of the routes will mostly be this app itself, especially with the providers being implemented.

lib/TextProcessing/HeadlineProvider.php
MB-Finski: If the userId is null, the quota checking is omitted currently. The other option would be to disable scheduling all together for now until your core pr is merged -- if the admin wishes to enforce a quota with absolute certainty.

lib/Db/QuotaUsage.php
MB-Finski: Done

appinfo/info.xml
MB-Finski: Reverted the version bump.

MB-Finski:
> Do you have screenshots of the front-end changes, before and after?
Changes were made only to the admin and personal settings pages, as follows.

Before:
image
image


Here's  after:
image
image

lib/Cron/CleanupQuotaDb.php
```diff
@@ -0,0 +1,60 @@
...
+
+class CleanupQuotaDb extends TimedJob {
+	public function __construct(
+		ITimeFactory $time,
+		private QuotaUsageMapper $quotaUsageMapper,
+		private LoggerInterface $logger,
+		private IConfig $config
+	) {
+		parent::__construct($time);
+		$this->setInterval(60 * 60 * 1); // Daily
```
marcelklehr, line 43: daily?

MB-Finski: Haha, oupsie! That was there for debugging...

lib/Cron/CleanupQuotaDb.php
```diff
@@ -0,0 +1,60 @@
...
+		$this->logger->debug('Run cleanup job for OpenAI quota db');
+		$this->quotaUsageMapper->cleanupQuotaUsages(
+			max(
+				intval($this->config->getAppValue(
+					Application::APP_ID,
+					'quota_period',
+					strval(Application::DEFAULT_QUOTA_PERIOD)
+				)),
+				Application::DEFAULT_QUOTA_PERIOD
+			)
```
marcelklehr, line 56: So, if quota_period is smaller than DEFAULT_QUOTA_PERIOD it will be ignored?

lib/Db/ImageGenerationMapper.php
```diff
@@ -163,6 +164,11 @@ public function cleanupGenerations(int $maxAge = Application::MAX_GENERATION_IDL
 
 		/** @var ImageGeneration[] $generations */
 		$generations = $this->findEntities($qb);
+		$qb->resetQueryParts();
```
marcelklehr, line 167: I usually create a new qb via `$qb = $this->db->getQueryBuilder();`, feels cleaner somehow

lib/Cron/CleanupQuotaDb.php
MB-Finski: This is an intentional compromise; if you set the quota period to 1 day for a while and then set the period back up to, say, 15 days you'll have lost the quota entries beyond 1 day (actually 1 day before the cron job ran).

MB-Finski: I'm kind of assuming that 30 days is the most used quota interval.

lib/Service/OpenAiSettingsService.php
```diff
@@ -0,0 +1,433 @@
...
+	/**
+	 * Set the admin config for the settings page
+	 * @param mixed[] $config
+	 * @return void
+	 * @throws Exception
+	 */
+	public function setAdminConfig(array $adminConfig): void {
+		// That the variable types are correct
+		foreach (array_keys($adminConfig) as $key) {
+			if (gettype($adminConfig[$key]) !== self::ADMIN_CONFIG_TYPES[$key]) {
```
marcelklehr, line 321: Cool idea!

lib/TextProcessing/FreePromptProvider.php
```diff
@@ -27,14 +29,17 @@ public function getName(): string {
...
-		if (isset($completion['choices']) && is_array($completion['choices']) && count($completion['choices']) > 0) {
-			$choice = $completion['choices'][0];
-			if (isset($choice['message'], $choice['message']['content'])) {
-				return $choice['message']['content'];
-			}
+		// Max tokens are limited later to max tokens specified in the admin settings so here we just request PHP_INT_MAX
+		try {
+			$completion = $this->openAiAPIService->createChatCompletion($this->userId, $prompt, 1, $adminModel, PHP_INT_MAX, false);
+		} catch (Exception $e) {
+			throw new Exception('OpenAI/LocalAI request failed: ' . $e->getMessage());
```
marcelklehr, line 36: The TextProcessing API expects a RuntimeException I believe

lib/TextProcessing/FreePromptProvider.php
```diff
@@ -27,14 +29,17 @@ public function getName(): string {
...
 		}
-		throw new \Exception('No result in OpenAI/LocalAI response. ' . ($completion['error'] ?? ''));
+		if (count($completion) > 0) {
+			return array_pop($completion);
+		}
+
+		throw new Exception('No result in OpenAI/LocalAI response. ' . ($completion['error'] ?? ''));
```
marcelklehr, line 42: same here

lib/TextProcessing/HeadlineProvider.php
marcelklehr: > If the userId is null, the quota checking is omitted currently.

Cool! Just wanted to make sure you know that userId is most likely not set

lib/TextProcessing/HeadlineProvider.php
```diff
@@ -28,14 +30,17 @@ public function getName(): string {
...
-		if (isset($completion['choices']) && is_array($completion['choices']) && count($completion['choices']) > 0) {
-			$choice = $completion['choices'][0];
-			if (isset($choice['message'], $choice['message']['content'])) {
-				return $choice['message']['content'];
-			}
+		// Max tokens are limited later to max tokens specified in the admin settings so here we just request PHP_INT_MAX
+		try {
+			$completion = $this->openAiAPIService->createChatCompletion($this->userId, $prompt, 1, $adminModel, PHP_INT_MAX, false);
+		} catch (Exception $e) {
+			throw new Exception('OpenAI/LocalAI request failed: ' . $e->getMessage());
```
marcelklehr, line 37: Should be a RuntimeException I think

marcelklehr:
Frontend looks really nice :blue_heart:

lib/Cron/CleanupQuotaDb.php
marcelklehr: Ah, that's neat. maybe a code comment is on order here, so future readers know what this is about :)

lib/Controller/OpenAiAPIController.php
```diff
@@ -147,14 +164,65 @@ public function getImageGenerationContent(string $hash, int $urlId): DataDisplay
...
+	 * @return DataResponse
+	 */
+	#[NoAdminRequired]
+	public function clearPromptHistory(?bool $clearTextPrompts = null, ?bool $clearImagePrompts = null): DataResponse {
+		$this->logger->debug('Clearing prompt history: ' . 'clearTextPrompts ' . strval($clearTextPrompts) . ' clearImagePrompts ' . strval($clearImagePrompts));
+		if ($clearTextPrompts === true) {
+			try {
+				$this->openAiAPIService->clearPromptHistory($this->userId, Application::PROMPT_TYPE_TEXT);
+			} catch (DBException $e) {
+				return new DataResponse($this->l10n->t('Unknown error while clearing prompt history'), Http::STATUS_BAD_REQUEST);
```
marcelklehr, line 192: I guess one could take the stance that it doesn't matter what error response code you use as long as it's an error code. Still, semantically the client did nothing wrong if the database fails, which would make it a 5xx status. I'm not saying you should make sure to always get the most specific status code you can find, but distinguishing 500 vs 400 would be nice IMHO :shrug:

MB-Finski: Sounds sensible, I'll change the error handling such that it returns 5xx for server side failures.

lib/Db/ImageGenerationMapper.php
MB-Finski: Thanks! Done!

lib/TextProcessing/FreePromptProvider.php
MB-Finski: Thanks, it's fixed!

lib/TextProcessing/FreePromptProvider.php
MB-Finski: Thanks, it's fixed!

lib/TextProcessing/HeadlineProvider.php
MB-Finski: Thanks, it's fixed!

lib/Controller/OpenAiAPIController.php
MB-Finski: Done: https://github.com/nextcloud/integration_openai/pull/54/commits/630648b4d240eefce15c27ef9ad6153acd5928f8

lib/Cron/CleanupQuotaDb.php
MB-Finski: Comment added!

marcelklehr (approved):
LGTM

Issue #60, opened 2023-11-12
ChatGPT-Like Answer not possible anymore


manuelkamp:
### Which version of integration_openai are you using?

1.1.1

### Which version of Nextcloud are you using?

27.1.3

### Which browser are you using? In case you are using the phone App, specify the Android or iOS version and device please.

Edge, Firefox latest (not relevant)

### Describe the Bug

"Completion request error: Unknown text generation error" Error shows up every time i want to use ChatGPT-Like Answers through my LocalAI. LocalAI itself is working fine, tried it via ssh-curl on my local LocalAI instance and on my remote nextcloud instance. The issue has to be this plugin. I did not change any settings since I set it up last week. However i saw that it did upgrade the app (automatically) from 1.0.13 to 1.1.1.
However, Image generation via my LocalAI works fine via this app.

### Expected Behavior

Get a response from LocalAI

### To Reproduce

Open textfile, type / and select ChatGPT-Like answer and type anything in it. it results in the error shown in the right upper corner.
Error in logfile:
{"reqId":"ZVDYoqeILvOTuLnCwMkWEQAAAAI","level":2,"time":"2023-11-12T13:52:34+00:00","remoteAddr":"10.0.0.1","user":"manuel","app":"integration_openai","method":"POST","url":"/apps/integration_openai/completions","message":"Text generation error: {\"created\":1698852566,\"object\":\"text_completion\",\"id\":\"b36ddd74-a267-412b-b50d-7a50865e5447\",\"model\":\"nous-hermes-llama2-13b\",\"usage\":{\"prompt_tokens\":0,\"completion_tokens\":0,\"total_tokens\":0}}","userAgent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Safari/537.36 Edg/119.0.0.0","version":"27.1.3.2","data":{"app":"integration_openai"},"id":"6550da68ef49e"}

MB-Finski:
Thanks a ton for the report! I found the issue causing the request to the completions endpoint to be broken. I'll get the fix published ASAP. In the meanwhile you can bypass this by tricking the front-end to send the requests to the completions/chat endpoint by having the model name start with 'gpt-'.

MB-Finski:
But then again, maybe we should use the chat/completions endpoint by default since this will automatically inject the proper prompt template for the model in the llm backend which would improve the answer quality for most models. Hmm...

[...]

PR #61, opened 2023-11-12
Add admin setting for choosing endpoint when using LocalAI

MB-Finski:
Exposes choosing the endpoint for the LocalAI as an admin setting. This should improve text generations quite a bit since the chat endpoint will inject the proper model specific prompt template at the LocalAI end.

Also, fixes the params for calling the plain "completions" endpoint: https://github.com/nextcloud/integration_openai/issues/60

### DIFF
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -300,7 +300,7 @@ class OpenAiAPIService {
 		if (isset($response['usage'])) {
 			$usage = $response['usage']['total_tokens'];
 			try {
-				$this->quotaUsageMapper->createQuotaUsage($userId, Application::QUOTA_TYPE_TEXT, $usage);
+				$this->quotaUsageMapper->createQuotaUsage($userId ?? '', Application::QUOTA_TYPE_TEXT, $usage);
 			} catch (DBException $e) {
 				$this->logger->warning('Could not create quota usage for user: ' . $userId . ' and quota type: ' . Application::QUOTA_TYPE_TEXT . '. Error: ' . $e->getMessage(), ['app' => Application::APP_ID]);
 			}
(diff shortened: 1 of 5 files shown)


MB-Finski:
Addition to admin settings page:
image

lib/Service/OpenAiAPIService.php
```diff
@@ -367,7 +372,7 @@ public function createChatCompletion(
 			$usage = $response['usage']['total_tokens'];
 
 			try {
-				$this->quotaUsageMapper->createQuotaUsage($userId, Application::QUOTA_TYPE_TEXT, $usage);
+				$this->quotaUsageMapper->createQuotaUsage($userId ?? '', Application::QUOTA_TYPE_TEXT, $usage);
 			} catch (DBException $e) {
 				$this->logger->warning('Could not create quota usage for user: ' . $userId . ' and quota type: ' . Application::QUOTA_TYPE_TEXT . '. Error: ' . $e->getMessage(), ['app' => Application::APP_ID]);
```
kyteinsky, line 377: I think you should return from the function if the userId is null since it will not be called from a command at any point in time apparently. And maybe return after catch as well? This seems very permissive in terms of quota handling, or maybe it is intended to be?

src/components/PersonalSettings.vue
```diff
@@ -43,11 +43,11 @@
 					{{ t('integration_openai', 'Clear prompt history') }}
 				</label>
 				<button id="clear-text-prompt-history"
-					@click="clearPromptHistory(false,true)">
+					@click="clearPromptHistory(true, false)">
 					{{ t('integration_openai', 'Clear text prompts') }}
 				</button>
 				<button id="clear-image-prompt-history"
-					@click="clearPromptHistory(true,false)">
+					@click="clearPromptHistory(false, true)">
```
kyteinsky, line 50: original was correct

`clearPromptHistory(clearImages, clearText) {`

lib/Service/OpenAiAPIService.php
MB-Finski: Currently, if a task is scheduled to run in the background it will not have an associated session and userId would be null. Thus, this ensures that the endpoint cannot be overused by scheduling a task through OCS if quota is enforced. There is a PR for core by Marcel that provides a userId even for scheduled tasks but it's still WIP.

src/components/PersonalSettings.vue
MB-Finski: Haha, yeah. I just changed the boolean order to give the correct UI behaviour.

MB-Finski: Fixed in: https://github.com/nextcloud/integration_openai/pull/61/commits/d68521b73536713ed5472c93fc0f66867e29bfff

lib/Service/OpenAiAPIService.php
kyteinsky: Ah, thanks for the explanation!

kyteinsky (approved):
lgtm :balloon::balloon:

PR #66, opened 2023-12-06
Implement IProviderWithExpectedRuntime and IProviderWithUserId TP providers

julien-nc:
* Bump min NC version to 28 (when the new interfaces appeared
* Bump max NC version to 29 (not necessary but why not)
* Set an expected time of 10 seconds if OpenAI is used and 5 minutes if LocalAI is used

@marcelklehr Is it fine if `setUserId` overwrites the injected value in the class attribute?

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -71,7 +71,7 @@ Negative:
 
 Positive:
 * the software for training and inferencing of this model is open source
-* the trained model is freely available, and thus can be ran on-premises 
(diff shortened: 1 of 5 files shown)


marcelklehr (approved):

marcelklehr:
> @marcelklehr Is it fine if setUserId overwrites the injected value in the class attribute?

Yep, that should be ok. It's a bit unorthodox, but it works.

PR #69, opened 2023-12-17
Allow local IP address for LocalAI

jkellerer:
So far, to use LocalAI, one had to set `"allow_local_remote_servers": true` in `config.php`. 

However, I think this should not be a requirement for a service that is meant to run in the local network?

This PR adjusts the request to allow local addresses when using LocalAI, without a need to allow requests to local addresses for all other apps.

### DIFF
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -663,6 +663,10 @@ class OpenAiAPIService {
 				}
 			}
 
+			if (!$this->isUsingOpenAi()) {
+				$options["nextcloud"]["allow_local_address"] = true;
+			}
+			
 			if ($contentType === null) {
 				$options['headers']['Content-Type'] = 'application/json';
 			} elseif ($contentType === 'multipart/form-data') {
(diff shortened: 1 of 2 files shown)


bigcat88 (approved):
that is correct, imho(not tested myself)

jkellerer:
Guess the test failures are caused by modified `$options` not matching mock conditions required to return the expected response. Had not the full env setup to run the tests. Will update the PR.

MB-Finski:
Thanks for the pr (and fixing the tests, even). LGTM, but I'd still very much like to test it myself. I tend to isolate all my web servers to their own subnets (even my dev setup) so I've never had to fiddle with this setting myself. I'll have to setup an API wrapper service in the same subnet to test this; I'll see if I can get to it today or tomorrow!

jkellerer:
Btw. same subnet is not the primary trigger, in my case it was 2 different private networks and it still didn't work without this change.

MB-Finski:
> ary trigger, in my case it was 2 different private networks and it still didn't work without this change.

Ahh... So digging around the server source, the trigger (for IPs) to be considered "local" would be these two CIDRs: 100.64.0.0/10 and 192.0.0.0/24. Why it's like that in server core, I really don't know (esp. since there are various other ranges as well that are considered reserved for private network use). Any whoo: Tested the pr and it's working as expected working. Thanks!

PR #70, opened 2023-12-20
Implement a TextToImageProvider

julien-nc:
* We could adjust the expected runtime, not sure about the current values
* the request logic has been extracted from `OpenAiAPIService::createImage` into another method. When running a TextToImage task, we want to trigger the quota stuff but not the image generation mapper stuff as this is done by the assistant


MB-Finski (approved):
LGTM; Thanks for this! I think I'll add the low-pass filtering of the expected runtime in one go for both text processing and image-to-text so they can be as they are right now.

PR #71, opened 2023-12-22
Use lp-filtering to estimate expected runtime

MB-Finski:
Adds low-pass filtering for the expected runtime based on observed runtimes so that the consumers of the providers can have a more accurate estimate of the runtime.

### DIFF
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -639,12 +639,12 @@ class OpenAiAPIService {
 	 */
 	public function updateExpTextProcessingTime(int $runtime): void {
 		$oldTime = $this->getExpTextProcessingTime();
-		$newTime = intval((1 - Application::EXPECTED_RUNTIME_LOWPASS_FACTOR) * $oldTime + Application::EXPECTED_RUNTIME_LOWPASS_FACTOR * $runtime);
+		$newTime = (1 - Application::EXPECTED_RUNTIME_LOWPASS_FACTOR) * $oldTime + Application::EXPECTED_RUNTIME_LOWPASS_FACTOR * $runtime;
 
 		if ($this->isUsingOpenAi()) {
-			$this->config->setAppValue(Application::APP_ID, 'openai_text_generation_time', strval($newTime));
+			$this->config->setAppValue(Application::APP_ID, 'openai_text_generation_time', strval(intval($newTime)));
 		} else {
-			$this->config->setAppValue(Application::APP_ID, 'localai_text_generation_time', strval($newTime));
+			$this->config->setAppValue(Application::APP_ID, 'localai_text_generation_time', strval(intval($newTime)));
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


lib/Service/OpenAiAPIService.php
```diff
@@ -624,6 +624,54 @@ public function getGenerationImage(string $hash, int $urlId): ?array {
...
+	 * @return void
+	 */
+	public function updateExpTextProcessingTime(int $runtime): void {
+		$oldTime = $this->getExpTextProcessingTime();
+		$newTime = intval((1 - Application::EXPECTED_RUNTIME_LOWPASS_FACTOR) * $oldTime + Application::EXPECTED_RUNTIME_LOWPASS_FACTOR * $runtime);
+
+		if ($this->isUsingOpenAi()) {
+			$this->config->setAppValue(Application::APP_ID, 'openai_text_generation_time', strval($newTime));
+		} else {
+			$this->config->setAppValue(Application::APP_ID, 'localai_text_generation_time', strval($newTime));
```
kyteinsky, line 647: intval seems redundant, `newTime` can directly do `strval(...)`

kyteinsky: same in `updateExpImgProcessingTime`

lib/TextProcessing/FreePromptProvider.php
```diff
@@ -44,9 +46,11 @@ public function process(string $prompt): string {
 			throw new RunTimeException('OpenAI/LocalAI request failed: ' . $e->getMessage());
 		}
 		if (count($completion) > 0) {
+			$endTime = time();
+			$this->openAiAPIService->updateExpTextProcessingTime($endTime - $startTime);
 			return array_pop($completion);
 		}
-
+
```
kyteinsky, line 53: trailing spaces

kyteinsky:
lgtm other than that! Nice expected time adjustments!

lib/TextProcessing/FreePromptProvider.php
MB-Finski: I wondered why cs:fix didn't want to remove those. I'll push a fix!

lib/Service/OpenAiAPIService.php
MB-Finski: That's true. I just thought that it'd be nicer to store an integer instead of a float since we don't care much for a second here or there. But I'm fine with taking that out if you think it makes more sense.

kyteinsky: ah I didn't notice the float to int conversion, you can ignore this or do both operations above itself.

kyteinsky (approved):

lib/Service/OpenAiAPIService.php
MB-Finski: > or you could leave it as int even: https://github.com/nextcloud/server/blob/master/lib/public/IConfig.php#L136

Oh, I'd forgotten you could do that! Thanks!

PR #84, opened 2024-03-25
Fix: use basic auth to get generated images if necessary

julien-nc:
closes https://github.com/nextcloud/text2image_helper/issues/49

Auth is not an issue with OpenAI as result image URLs are accessible publicly.
But when LocalAI is set up with basic auth, we still don't use any auth when getting result images.
Tested with LocalAI behind a reverse proxy adding basic auth.

### DIFF
--- a/lib/TextToImage/TextToImageProvider.php
+++ b/lib/TextToImage/TextToImageProvider.php
@@ -8,6 +8,7 @@ namespace OCA\OpenAi\TextToImage;
 
 use OCA\OpenAi\AppInfo\Application;
 use OCA\OpenAi\Service\OpenAiAPIService;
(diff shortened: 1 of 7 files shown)


marcelklehr (approved):

lib/TextToImage/TextToImageProvider.php
```diff
@@ -17,6 +18,7 @@
 class TextToImageProvider implements IProvider {
 	public function __construct(
 		private OpenAiAPIService $openAiAPIService,
+		private OpenAiSettingsService $openAiSettingsService,
```
marcelklehr, line 21: Unecessary?

marcelklehr (approved):

PR #101, opened 2024-05-29
Fix incorrect numeric settings handling on the server side

julien-nc:
Arbitrary choice to use null to know when we should delete a setting value.
`parseInt` returns `NaN` with invalid values, this becomes `null` in the network request.
This should be pretty safe.


kyteinsky (approved):
:sparkles:

PR #103, opened 2024-05-30
Move personal and admin settings to AI sections

julien-nc:
Which makes more sense than "connected accounts".


kyteinsky (approved):
:sparkles:

PR #104, opened 2024-06-07
feat: Allow answering in the same langauge as the input

marcelklehr:
Works with popular models >8B. Llama 3 8B fails sometimes, sadly.


kyteinsky:
reformulate and summary's prompts seem to cater to completion models. Shall we change them for chat completions since non-chat completion is legacy as per openai?

marcelklehr:
> reformulate and summary's prompts seem to cater to completion models.


That was not the goal, but was the only way to get llama2 8B to understand the prompts correctly.

julien-nc (approved):

PR #102, opened 2024-05-29
Migrate to the new task processing API

julien-nc:
* [x] Migrate free prompt
* [x] Migrate summarize
* [x] Migrate headline
* [x] Create topics
* [x] Migrate image generation
* [x] Migrate speech to text
* [x] Implement a provider for the new ContextWrite task type
* [x] Implement a provider for the new Chat task type

Ready to merge IMO.

PR #110, opened 2024-07-17
Prepare 3.0.0

julien-nc:
### Added

- Create all task processing providers that were text processing ones + image generation + audio transcription @julien-nc [#102](https://github.com/nextcloud/integration_openai/pull/102)
- ContextWrite provider
- Chat provider
- Add password confirmation when setting sensitive values in user and admin settings

### Changed

- Drop old image generation, audio transcription and text processing providers
- Use nc/vue 8.14.0

### Fixed

- Fix topics provider output: make it comma separated


marcelklehr:
:tada:

PR #111, opened 2024-07-25
Adjust providers to changes in server

julien-nc:
And add an enum slot for the text2text provider to select the model.


lib/TaskProcessing/ContextWriteProvider.php
```diff
@@ -53,10 +61,26 @@ public function getOptionalInputShape(): array {
...
+		return [];
+	}
+
 	public function getOptionalOutputShape(): array {
 		return [];
 	}
 
+	public function getOptionalOutputShapeEnumValues(): array {
+		return [];
+	}
```
marcelklehr, line 82: :sweat_smile:

PR #118, opened 2024-08-11
fix #117

### LINKED ISSUE #117: Error generating images with LocalAI integrated with Nextcloud AI (CPU only)

2.0.3

29.0.4

_No response_

### Describe the Bug

I am encountering an error while generating images using LocalAI integrated with Nextcloud AI. The error appears in the LocalAI logs as follows:

```
ERR Server error error="could not load model: rpc error: code = Unknown desc = stat /build/models/stablediffusion: no such file or directory" ip=172.18.1.2 latency=2.004227305s method=POST status=500 url=/v1/images/generations
```

However, when I use the LocalAI web interface directly, image generation works fine.

### Expected Behavior

Image generated with Local AI

### To Reproduce

1. Integrate LocalAI with Nextcloud AI.
2. Attempt to generate an image through the Nextcloud AI interface.

manuelkamp:
for usage with model "stablediffusion-cpp" in localai

### DIFF
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -525,6 +525,7 @@ class OpenAiAPIService {
 		$params = [
 			'prompt' => $prompt,
 			'size' => $size,
+			'model' => 'stablediffusion-cpp',
 		];
 
 		if ($this->isUsingOpenAi()) {


DaphneMuller:
hi, I'm sorry for the slow review. I've asked the maintainer of this app for an update here

kyteinsky:
Hi @manuelkamp, we can't hardcode the model name in code since that would break it for everyone who happens to use models with names other than "stablediffusion-cpp". See #122 PR that adds a model dropdown feature for image models which should fix your issue.

PR #135, opened 2024-10-01
Add admin setting to set the default image size

julien-nc:
And show the default size in the label for the optional task type's input field.


kyteinsky (approved):
:rocket:

PR #136, opened 2024-10-03
Encrypt api keys and basic passwords for app config and user settings

julien-nc:
And migrate the existing values.

### DIFF
--- a/lib/Migration/Version030102Date20241003155512.php
+++ b/lib/Migration/Version030102Date20241003155512.php
@@ -1,0 +1,89 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2020 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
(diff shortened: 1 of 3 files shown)


lib/Migration/Version030102Date20241003155512.php
```diff
@@ -0,0 +1,89 @@
...
+		)
+			->andWhere(
+				$qbSelect->expr()->isNotNull('configvalue')
+			);
+		$req = $qbSelect->executeQuery();
+		while ($row = $req->fetch()) {
+			$userId = $row['userid'];
+			$configKey = $row['configkey'];
+			$storedClearValue = $row['configvalue'];
+			$encryptedValue = $this->crypto->encrypt($storedClearValue);
```
marcelklehr, line 81: Do we need to check here if the value is `''` before encrypting?

marcelklehr (approved):

lib/Migration/Version030102Date20241003155512.php
julien-nc: No because `$qbSelect->expr()->nonEmptyString('configvalue')`.

PR #139, opened 2024-10-09
Switch to IAppConfig

julien-nc:
So we can mark some values as sensitive and the encryption is handled by the core.

I left out the "old" providers because we can now get rid of them (as TaskProcessing providers can be used by the old APIs).

Also, I didn't use the getValueInt and setValueInt methods because I don't know how it behaves when the value stored in db is an empty string so I was afraid it would mess with the default values and too lazy to try and check how it behaves.

### DIFF
--- a/lib/Service/OpenAiSettingsService.php
+++ b/lib/Service/OpenAiSettingsService.php
@@ -25,6 +25,8 @@ namespace OCA\OpenAi\Service;
 
 use Exception;
 use OCA\OpenAi\AppInfo\Application;
+use OCP\Exceptions\AppConfigTypeConflictException;
+use OCP\IAppConfig;
 use OCP\IConfig;
 use OCP\PreConditionNotMetException;
 use OCP\Security\ICrypto;
@@ -61,6 +63,7 @@ class OpenAiSettingsService {
 
 	public function __construct(
 		private IConfig $config,
+		private IAppConfig $appConfig,
 		private ICrypto $crypto,
 	) {
 
@@ -74,8 +77,7 @@ class OpenAiSettingsService {
 	 * @throws Exception
 	 */
 	public function getAdminApiKey(): string {
-		$encryptedApiKey = $this->config->getAppValue(Application::APP_ID, 'api_key');
-		return $encryptedApiKey === '' ? '' : $this->crypto->decrypt($encryptedApiKey);
+		return $this->appConfig->getValueString(Application::APP_ID, 'api_key');
 	}
 
 	/**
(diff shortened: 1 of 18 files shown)


lib/Service/OpenAiSettingsService.php
```diff
@@ -99,78 +100,84 @@ public function getUserApiKey(?string $userId, bool $fallBackOnAdminValue = fals
...
-		$quotas = json_decode($this->config->getAppValue(Application::APP_ID, 'quotas', json_encode(Application::DEFAULT_QUOTAS)) ?: json_encode(Application::DEFAULT_QUOTAS), true);
+		$quotas = json_decode(
+			$this->appConfig->getValueString(
+				Application::APP_ID, 'quotas',
+				json_encode(Application::DEFAULT_QUOTAS)
+			) ?: json_encode(Application::DEFAULT_QUOTAS),
+			true,
+		);
 		// Make sure all quota types are set in the json encoded app value (in case new quota types are added in the future)
 		if (count($quotas) !== count(Application::DEFAULT_QUOTAS)) {
```
kyteinsky, line 174: null check missing on json_decode's output

lib/Service/OpenAiSettingsService.php
```diff
@@ -180,7 +187,7 @@ public function getQuotas(): array {
 	 * @return boolean
 	 */
 	public function getChatEndpointEnabled(): bool {
-		return $this->config->getAppValue(Application::APP_ID, 'chat_endpoint_enabled', '0') === '1';
+		return $this->appConfig->getValueString(Application::APP_ID, 'chat_endpoint_enabled', '0') === '1';
```
kyteinsky, line 190: let's default to chat endpoint since many services assume so and it is more usable in general? What do you say?

lib/Service/OpenAiSettingsService.php
```diff
@@ -343,12 +351,7 @@ public function setQuotas(array $quotas): void {
 	 */
 	public function setAdminApiKey(string $apiKey): void {
 		// No need to validate. As long as it's a string, we're happy campers
-		if ($apiKey === '') {
-			$this->config->setAppValue(Application::APP_ID, 'api_key', '');
-		} else {
-			$encryptedApiKey = $this->crypto->encrypt($apiKey);
-			$this->config->setAppValue(Application::APP_ID, 'api_key', $encryptedApiKey);
-		}
+		$this->appConfig->setValueString(Application::APP_ID, 'api_key', $apiKey, false, true);
```
kyteinsky, line 354: this needs to be encrypted probably

lib/Service/OpenAiSettingsService.php
julien-nc: There shouldn't be any case where the stored value is not valid JSON but let's be careful, I agree.

lib/Service/OpenAiSettingsService.php
```diff
@@ -334,7 +342,7 @@ public function setQuotas(array $quotas): void {
 			}
 		}
 
-		$this->config->setAppValue(Application::APP_ID, 'quotas', json_encode($quotas));
+		$this->appConfig->setValueString(Application::APP_ID, 'quotas', json_encode($quotas));
```
kyteinsky, line 345: we can add a flag for `json_encode` to throw here or should do a `false` check, It is fine in other places but this might return false.

lib/Service/OpenAiSettingsService.php
julien-nc: the 4th param of appConfig->setValueString is "sensitive". Setting it to true encrypts the value.

kyteinsky: saw the PR description again, and very cool!

lib/Service/OpenAiSettingsService.php
julien-nc: :+1:

kyteinsky (approved):

lib/Service/OpenAiSettingsService.php
kyteinsky: sure

PR #143, opened 2024-10-11
feat: add ability to customize API root path in service URL

### LINKED ISSUE #105: Model selection field does not show up when trying to use perplexity openai endpoint

2.0.1

v29.0.2

126.0.1 (64-bit)

### Describe the Bug

I try to configure the openai endpoint of perplexity, bit the model selection just does not show up.
nextcloud_perplexity

### Expected Behavior

I can select the model.

### To Reproduce

Try to use perplexity as ai endpoint.

edward-ly:
Resolves #105.

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -19,7 +19,7 @@
 						<EarthIcon />
 					</NcTextField>
 					<NcButton type="tertiary"
-						:title="t('integration_openai', 'Leave empty to use {openaiApiUrl}', { openaiApiUrl: 'https://api.openai.com' })">
+						:title="t('integration_openai', 'Leave empty to use {openaiApiUrl}', { openaiApiUrl: 'https://api.openai.com/v1' })">
 						<template #icon>
 							<HelpCircleIcon />
 						</template>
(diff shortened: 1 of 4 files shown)


src/components/AdminSettings.vue
```diff
@@ -19,15 +19,15 @@
 						<EarthIcon />
 					</NcTextField>
 					<NcButton type="tertiary"
-						:title="t('integration_openai', 'Leave empty to use {openaiApiUrl}', { openaiApiUrl: 'https://api.openai.com' })">
+						:title="t('integration_openai', 'Leave empty to use {openaiApiUrl}', { openaiApiUrl: 'https://api.openai.com/v1' })">
```
kyteinsky, line 22: let's add v1 to line 15 too

src/components/AdminSettings.vue
```diff
@@ -19,15 +19,15 @@
...
 						</template>
 					</NcButton>
 				</div>
 				<NcNoteCard type="info">
 					<p>
-						{{ t('integration_openai', 'This should be the address of your LocalAI instance (or any service implementing an API similar to OpenAI). This URL will be accessed by your Nextcloud server.') }}
+						{{ t('integration_openai', 'This should include the address of your LocalAI instance (or any service implementing an API similar to OpenAI) along with the root path of the API. This URL will be accessed by your Nextcloud server.') }}
 					</p>
 					<p>
 						{{ t('integration_openai', 'This can be a local address with a port like {example}. In this case make sure \'allow_local_remote_servers\' is set to true in config.php', { example : 'http://localhost:8080' }) }}
```
kyteinsky, line 33: same here, v1 should be appended to the url

kyteinsky:
Adding a preview like a computed URL for an example endpoint like chat completions `http://localhost:8080/v1/chat/completions` when the service url field is modified would be convenient as Julien said here: https://github.com/nextcloud/integration_openai/issues/105#issuecomment-2396639019

julien-nc (changes requested):
I agree with the review from @kyteinsky . Appart from that, all good.
For appconfig values, using IAppConfig in the migration step was the right thing to do :+1:

edward-ly:
Done. I decided to use another NcNoteCard to display the preview as I thought NcTextField's helperText in dark mode was too faint to be usable, but I am open to other ideas.

kyteinsky (approved):

julien-nc (approved):
:+1:

PR #157, opened 2024-11-19
feat: add input chunking to summarize task

### LINKED ISSUE #150: Chunking for summarize
Let's implement chunking in the same way it was done in LLM2 to allow summarizing texts that are longer than the model context size.

* Implement chunking (maybe have the chunking logic in a service so it can later be used by other providers like "extract topics")
* Make the chunk size configurable in the admin settings
* Test with GPT-3.5, GPT-4 and a few small models with LocalAI (or another self hosted service)

edward-ly:
Closes #150. Lightly tested with gpt-4o, gpt-4o-mini, and gpt-3.5-turbo.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -44,6 +44,8 @@ class Application extends App implements IBootstrap {
 	public const DEFAULT_TRANSCRIPTION_MODEL_ID = 'whisper-1';
 	public const DEFAULT_DEFAULT_IMAGE_SIZE = '512x512';
 	public const MAX_GENERATION_IDLE_TIME = 60 * 60 * 24 * 10;
+	public const DEFAULT_CHUNK_SIZE = 10000;
+	public const MIN_CHUNK_SIZE = 500;
 	public const DEFAULT_MAX_NUM_OF_TOKENS = 1000;
 	public const DEFAULT_QUOTA_PERIOD = 30;
 
--- a/lib/Service/OpenAiSettingsService.php
+++ b/lib/Service/OpenAiSettingsService.php
@@ -42,6 +42,7 @@ class OpenAiSettingsService {
 		'default_stt_model_id' => 'string',
 		'default_image_model_id' => 'string',
 		'default_image_size' => 'string',
+		'chunk_size' => 'integer',
 		'max_tokens' => 'integer',
 		'llm_extra_params' => 'string',
 		'quota_period' => 'integer',
@@ -151,6 +152,13 @@ class OpenAiSettingsService {
 	 */
 	public function getRequestTimeout(): int {
 		return intval($this->appConfig->getValueString(Application::APP_ID, 'request_timeout', strval(Application::OPENAI_DEFAULT_REQUEST_TIMEOUT))) ?: Application::OPENAI_DEFAULT_REQUEST_TIMEOUT;
+	}
+
+	/**
+	 * @return int
+	 */
+	public function getChunkSize(): int {
+		return $this->appConfig->getValueInt(Application::APP_ID, 'chunk_size', Application::DEFAULT_CHUNK_SIZE) ?: Application::DEFAULT_CHUNK_SIZE;
 	}
 
 	/**
@@ -275,6 +283,7 @@ class OpenAiSettingsService {
 			'default_stt_model_id' => $this->getAdminDefaultSttModelId(),
 			'default_image_model_id' => $this->getAdminDefaultImageModelId(),
 			'default_image_size' => $this->getAdminDefaultImageSize(),
+			'chunk_size' => strval($this->getChunkSize()),
 			'max_tokens' => $this->getMaxTokens(),
 			'llm_extra_params' => $this->getLlmExtraParams(),
 			// Updated to get max tokens
@@ -463,6 +472,20 @@ class OpenAiSettingsService {
 	}
 
 	/**
+	 * Setter for chunkSize; default/minimum is 0 (no chunking)
+	 * @param int $chunkSize
+	 * @return void
+	 */
+	public function setChunkSize(int $chunkSize): void {
+		// Validate input:
+		$chunkSize = max(0, $chunkSize);
+		if ($chunkSize) {
+			$chunkSize = max(Application::MIN_CHUNK_SIZE, $chunkSize);
+		}
+		$this->appConfig->setValueInt(Application::APP_ID, 'chunk_size', $chunkSize);
+	}
+
+	/**
 	 * Setter for maxTokens; minimum is 100
 	 * @param int $maxTokens
 	 * @return void
@@ -588,6 +611,9 @@ class OpenAiSettingsService {
 		}
 		if (isset($adminConfig['default_image_size'])) {
 			$this->setAdminDefaultImageSize($adminConfig['default_image_size']);
+		}
+		if (isset($adminConfig['chunk_size'])) {
+			$this->setChunkSize(intval($adminConfig['chunk_size']));
 		}
 		if (isset($adminConfig['max_tokens'])) {
 			$this->setMaxTokens($adminConfig['max_tokens']);
--- a/lib/TaskProcessing/SummaryProvider.php
+++ b/lib/TaskProcessing/SummaryProvider.php
@@ -108,10 +108,9 @@ class SummaryProvider implements ISynchronousProvider {
 			$maxTokens = $input['max_tokens'];
 		}
 
+		$model = $this->openAiSettingsService->getAdminDefaultCompletionModelId();
 		if (isset($input['model']) && is_string($input['model'])) {
 			$model = $input['model'];
-		} else {
-			$model = $this->appConfig->getValueString(Application::APP_ID, 'default_completion_model_id', Application::DEFAULT_MODEL_ID) ?: Application::DEFAULT_MODEL_ID;
 		}
 
 		try {
--- a/lib/TaskProcessing/SummaryProvider.php
+++ b/lib/TaskProcessing/SummaryProvider.php
@@ -121,12 +121,18 @@ class SummaryProvider implements ISynchronousProvider {
 			try {
 				$completions = [];
 				if ($this->openAiAPIService->isUsingOpenAi() || $this->openAiSettingsService->getChatEndpointEnabled()) {
+					$summarySystemPrompt = 'Summarize the following text in the same language as the text.';
+
 					foreach ($prompts as $p) {
-						$completion = $this->openAiAPIService->createChatCompletion($userId, $model, $p, null, null, 1, $maxTokens);
+						$completion = $this->openAiAPIService->createChatCompletion($userId, $model, $p, $summarySystemPrompt, null, 1, $maxTokens);
 						$completions[] = $completion['messages'];
 					}
 				} else {
-					foreach ($prompts as $p) {
+					$wrapSummaryPrompt = function (string $p): string {
+						return 'Summarize the following text delimited by XML tags in the same language as the text.\n\n<text>\n' . $p . '\n</text>';
+					};
+
+					foreach (array_map($wrapSummaryPrompt, $prompts) as $p) {
 						$completions[] = $this->openAiAPIService->createCompletion($userId, $p, 1, $model, $maxTokens);
 					}
 				}
@@ -161,13 +167,9 @@ class SummaryProvider implements ISynchronousProvider {
 		// It's safer to have a lower estimate on the max number of tokens, so consider 3 bytes per token instead of 4 (to account for some multibyte characters)
 		$maxChars = $chunkSize * 3;
 
-		$wrapSummaryPrompt = function (string $p): string {
-			return 'Summarize the following text. Detect the language of the text. Use the same language as the text.  Output only the summary.  Here is the text:' . "\n\n" . $p . "\n\n" . 'Here is your summary in the same language as the text:';
-		};
-
 		if (!$chunkSize || (mb_strlen($prompt) <= $maxChars)) {
 			// Chunking is disabled or prompt is short enough to be a single chunk
-			return [$wrapSummaryPrompt($prompt)];
+			return [$prompt];
 		}
 
 		// Try splitting by paragraph, match as many paragraphs as possible per chunk up to the maximum chunk size
--- a/lib/TaskProcessing/SummaryProvider.php
+++ b/lib/TaskProcessing/SummaryProvider.php
@@ -101,9 +101,8 @@ class SummaryProvider implements ISynchronousProvider {
 			throw new RuntimeException('Invalid prompt');
 		}
 		$prompt = $input['input'];
-		$prompt = 'Summarize the following text. Detect the language of the text. Use the same language as the text.  Output only the summary.  Here is the text:' . "\n\n" . $prompt . "\n\n" . 'Here is your summary in the same language as the text:';
 
-		$maxTokens = null;
+		$maxTokens = $this->openAiSettingsService->getMaxTokens();
 		if (isset($input['max_tokens']) && is_int($input['max_tokens'])) {
 			$maxTokens = $input['max_tokens'];
 		}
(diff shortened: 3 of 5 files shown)


lib/TaskProcessing/SummaryProvider.php
```diff
@@ -101,7 +101,18 @@ public function process(?string $userId, array $input, callable $reportProgress)
...
+		$wrapSummaryPrompt = function (string $p): string {
+			return 'Summarize the following text. Detect the language of the text. Use the same language as the text.  Output only the summary.  Here is the text:' . "\n\n" . $p . "\n\n" . 'Here is your summary in the same language as the text:';
+		};
+
+		$chunkSize = $this->openAiSettingsService->getChunkSize();
+		$prompts = [$wrapSummaryPrompt($prompt)];
+		if ($chunkSize) {
+			$prompts = explode('<|split|>', chunk_split($prompt, $chunkSize, '<|split|>'));
+			$prompts = array_map($wrapSummaryPrompt, $prompts);
+			array_pop($prompts);
```
julien-nc, line 114:
```suggestion
			array_pop($prompts);
			$prompts = array_map($wrapSummaryPrompt, $prompts);
```

lib/TaskProcessing/SummaryProvider.php
```diff
@@ -115,18 +126,23 @@ public function process(?string $userId, array $input, callable $reportProgress)
...
 			}
 		} catch (Exception $e) {
 			throw new RuntimeException('OpenAI/LocalAI request failed: ' . $e->getMessage());
 		}
-		if (count($completion) > 0) {
+		if (array_reduce($completions, fn (bool $prev, array $next): bool => $prev && (count($next) > 0), true)) {
 			$endTime = time();
 			$this->openAiAPIService->updateExpTextProcessingTime($endTime - $startTime);
-			return ['output' => array_pop($completion)];
+			return ['output' => implode(' ', array_map('array_pop', $completions))];
```
julien-nc, line 145:
```suggestion
			$completionStrings = array_map('array_pop', $completions)
			return ['output' => implode(' ', $completionStrings)];
```
To fix the tests.

lib/Service/OpenAiSettingsService.php
```diff
@@ -461,6 +470,17 @@ public function setRequestTimeout(int $requestTimeout): void {
...
 
+	/**
+	 * Setter for chunkSize; default/minimum is 0 (no chunking)
+	 * @param int $chunkSize
+	 * @return void
+	 */
+	public function setChunkSize(int $chunkSize): void {
+		// Validate input:
+		$chunkSize = max(0, $chunkSize);
+		$this->appConfig->setValueString(Application::APP_ID, 'chunk_size', strval($chunkSize));
```
julien-nc, line 481: We could use setValueInt et getValueInt for the chunk size. I was just lazy when switching from IConfig to IAppConfig to use the type-specific methods and migrate the existing values. For a new one like chunk_size, we can go for the typed stuff IMO. (Optional change)

julien-nc (changes requested):
:+1:

The admin setting input could be moved in the "Text generation" section. Wdyt?

edward-ly:
> The admin setting input could be moved in the "Text generation" section. Wdyt?

Maybe the chunk size could also be considered as a usage limit, but since we only use it on the summarize task, I think it would make sense to move it there too.

lib/Service/OpenAiSettingsService.php
edward-ly: It works as long as the chunk_size string hasn't been written to the database yet. Trying to switch from string to int would result in a `AppConfigTypeConflictException: conflict with value type from database`.

lib/TaskProcessing/SummaryProvider.php
edward-ly: Even so, I found that `array_map` now emits a warning if the callback function expects an argument passed by reference, which is the case for `array_pop`. I'll probably need to refactor the function to indirectly call `array_pop` (and probably replace it with `end` as that might be faster).

https://www.php.net/manual/en/function.array-map.php#refsect1-function.array-map-changelog

lib/Service/OpenAiSettingsService.php
```diff
@@ -461,6 +470,17 @@ public function setRequestTimeout(int $requestTimeout): void {
...
 	}
 
+	/**
+	 * Setter for chunkSize; default/minimum is 0 (no chunking)
+	 * @param int $chunkSize
+	 * @return void
+	 */
+	public function setChunkSize(int $chunkSize): void {
+		// Validate input:
+		$chunkSize = max(0, $chunkSize);
```
marcelklehr, line 480: Very elegant :v:

lib/TaskProcessing/SummaryProvider.php
```diff
@@ -115,18 +126,26 @@ public function process(?string $userId, array $input, callable $reportProgress)
...
+				foreach ($prompts as $prompt) {
+					$completions[] = $this->openAiAPIService->createCompletion($userId, $prompt, 1, $model, $maxTokens);
+				}
 			}
 		} catch (Exception $e) {
 			throw new RuntimeException('OpenAI/LocalAI request failed: ' . $e->getMessage());
 		}
-		if (count($completion) > 0) {
+
+		$allPromptsHaveCompletions = array_reduce($completions, fn (bool $prev, array $next): bool => $prev && count($next), true);
```
marcelklehr, line 143: Elegant :blue_heart: Love it :)

marcelklehr (approved):
The code looks really good. The algorithm deviates a little from what llm2 does, though: In llm2 the summaries of the chunks are concatenated and fed through the same algorithm again (ie. chunked and summarized) until there is only one summary left.

marcelklehr (changes requested):
Oops, clicked the wrong button.

edward-ly:
> The code looks really good. The algorithm deviates a little from what llm2 does, though: In llm2 the summaries of the chunks are concatenated and fed through the same algorithm again (ie. chunked and summarized) until there is only one summary left.

Now that you mention it, I do see the similar loop in llm2 now, fixed.

lib/TaskProcessing/SummaryProvider.php
```diff
@@ -114,21 +119,40 @@ public function process(?string $userId, array $input, callable $reportProgress)
 			$model = $this->appConfig->getValueString(Application::APP_ID, 'default_completion_model_id', Application::DEFAULT_MODEL_ID) ?: Application::DEFAULT_MODEL_ID;
 		}
 
-		try {
-			if ($this->openAiAPIService->isUsingOpenAi() || $this->openAiSettingsService->getChatEndpointEnabled()) {
-				$completion = $this->openAiAPIService->createChatCompletion($userId, $model, $prompt, null, null, 1, $maxTokens);
-			} else {
-				$completion = $this->openAiAPIService->createCompletion($userId, $prompt, 1, $model, $maxTokens);
+		while (strlen($prompt) > $maxTokens) {
```
marcelklehr, line 122: Could this turn into an infinite loop?

edward-ly: If max_tokens is set small enough, I can see it being a possibility. Maybe whenever the output summary is longer than the input prompt, that could be another terminating condition?

marcelklehr: Intuitively, I'd say the stop condition should be "only one summary is left"

edward-ly: But if each chunk has their own summary, what does it mean to have "only one left/remaining"?

marcelklehr: > But if each chunk has their own summary, what does it mean to have "only one left/remaining"?

Since we expect the summaries to shorten the input, with each iteration there should be less chunks. So, I guess to have only one summary left, there should be only one chunk left.

edward-ly: Quick fix is done.

marcelklehr: Thanks! Works nicely.

marcelklehr:
I'm about to change the algorithm in llm2 slightly, so we should change it here as well a last time: When the input is shorter than the chunk size it should still go through the llm summarizer once.

lib/AppInfo/Application.php
```diff
@@ -41,6 +41,7 @@ class Application extends App implements IBootstrap {
 	public const DEFAULT_TRANSCRIPTION_MODEL_ID = 'whisper-1';
 	public const DEFAULT_DEFAULT_IMAGE_SIZE = '512x512';
 	public const MAX_GENERATION_IDLE_TIME = 60 * 60 * 24 * 10;
+	public const DEFAULT_CHUNK_SIZE = 0; // no chunking
```
marcelklehr, line 44: I wonder if we should add a sensible default here? GPT3.5 has a context size of 16385 for example. We could set this to 10k? What do you think?

lib/TaskProcessing/SummaryProvider.php
```diff
@@ -101,9 +101,14 @@ public function process(?string $userId, array $input, callable $reportProgress)
...
-		$prompt = 'Summarize the following text. Detect the language of the text. Use the same language as the text.  Output only the summary.  Here is the text:' . "\n\n" . $prompt . "\n\n" . 'Here is your summary in the same language as the text:';
 
-		$maxTokens = null;
+		$wrapSummaryPrompt = function (string $p): string {
+			return 'Summarize the following text. Detect the language of the text. Use the same language as the text.  Output only the summary.  Here is the text:' . "\n\n" . $p . "\n\n" . 'Here is your summary in the same language as the text:';
+		};
+
+		$chunkSize = $this->openAiSettingsService->getChunkSize();
+
+		$maxTokens = Application::DEFAULT_MAX_NUM_OF_TOKENS;
```
julien-nc, line 111: Off-topic: maxTokens should be initialized with the `max_tokens` config value.

src/components/AdminSettings.vue
```diff
@@ -162,6 +162,28 @@
 				<h2>
 					{{ t('integration_openai', 'Text generation') }}
 				</h2>
+				<div class="line">
```
julien-nc, line 165: This could be placed after the "extra completion model parameters".

src/components/AdminSettings.vue
```diff
@@ -162,6 +162,28 @@
...
+						:label="t('integration_openai', 'Max input length per request')"
+						:show-trailing-button="!!state.chunk_size"
+						@update:value="onInput()"
+						@trailing-button-click="state.chunk_size = '' ; onInput()">
+						<template #trailing-button-icon>
+							<CloseIcon :size="20" />
+						</template>
+					</NcInputField>
+					<NcButton type="tertiary"
+						:title="t('integration_openai', 'Split the prompt into chunks with each chunk being no more than the specified length (0 disables chunking)')">
```
julien-nc, line 181: We need to be more explicit here that the size is in characters.

lib/TaskProcessing/SummaryProvider.php
```diff
@@ -114,21 +119,45 @@ public function process(?string $userId, array $input, callable $reportProgress)
...
+				} else {
+					foreach ($prompts as $prompt) {
+						$completions[] = $this->openAiAPIService->createCompletion($userId, $prompt, 1, $model, $maxTokens);
+					}
+				}
+			} catch (Exception $e) {
+				throw new RuntimeException('OpenAI/LocalAI request failed: ' . $e->getMessage());
+			}
+
+			$allPromptsHaveCompletions = array_reduce($completions, fn (bool $prev, array $next): bool => $prev && count($next), true);
```
julien-nc, line 146: Could you add a comment explaining this reduce so it's easier to read later?

lib/TaskProcessing/SummaryProvider.php
julien-nc: > Since we expect the summaries to shorten the input

We don't know that for sure. We could add a stop condition: when the number of chunks does not change between 2 iterations.
Wdyt?

lib/TaskProcessing/SummaryProvider.php
```diff
@@ -114,21 +119,45 @@ public function process(?string $userId, array $input, callable $reportProgress)
 			$model = $this->appConfig->getValueString(Application::APP_ID, 'default_completion_model_id', Application::DEFAULT_MODEL_ID) ?: Application::DEFAULT_MODEL_ID;
 		}
 
-		try {
-			if ($this->openAiAPIService->isUsingOpenAi() || $this->openAiSettingsService->getChatEndpointEnabled()) {
-				$completion = $this->openAiAPIService->createChatCompletion($userId, $model, $prompt, null, null, 1, $maxTokens);
-			} else {
-				$completion = $this->openAiAPIService->createCompletion($userId, $prompt, 1, $model, $maxTokens);
+		while (strlen($prompt) > $maxTokens) {
```
julien-nc, line 122: Could you add a comment here explaining the chunking logic in natural language?

lib/TaskProcessing/SummaryProvider.php
marcelklehr: I still see two issues here:

1.`$maxTokens` counts in tokens, while strlen counts in bytes
2. `$maxTokens` is the maximum amount of output tokens that we ask the model to produce. Why are we using it here to compare it to the length of the input?

lib/TaskProcessing/SummaryProvider.php
```diff
@@ -114,21 +119,45 @@ public function process(?string $userId, array $input, callable $reportProgress)
...
+			$prompts = [$wrapSummaryPrompt($prompt)];
+			$promptIsChunked = $chunkSize && (strlen($prompt) > $chunkSize);
+			if ($promptIsChunked) {
+				$prompts = explode('<|split|>', chunk_split($prompt, $chunkSize, '<|split|>'));
```
marcelklehr, line 126: Does `chunk_split` split along sentence boundaries? AFAIK it splits cleanly after x characters/bytes. This will break sentences at the boundaries of chunks. Can we implement a proper sentence boundary splitter?

marcelklehr: Additionally, I think the chunk size should be measured in tokens not in characters, so that admins can easily set the value from a models context size which is usually given in tokens.

lib/AppInfo/Application.php
marcelklehr: (See below about discrepancy between character count and token count.)

lib/TaskProcessing/SummaryProvider.php
marcelklehr: (One word is approximately 1.5 tokens on average, I believe.)

kyteinsky: > Can we implement a proper sentence boundary splitter?

:100:

exploding on `newlines -> fullstops and other delimiters -> words`
would be efficient in packing the chunks for a given chunk size, see what fits best, maybe the delimiter split will be troubling in blocks of code.

tokens can vary based on languages, see the token ratio section here: https://gptforwork.com/guides/openai-gpt3-tokens. Maybe we can relax it a little more like 2 to 2.5 tokens, considering multilingual text and occasional code blocks?

-- or, we can call a local GPT2 fast tokenizer to get the exact tokens, like we do in context chat backend (implicitly).

marcelklehr: 2.5 sounds good, we can overestimate, but we shouldn't underestimate

edward-ly: Why not use a similar algorithm that langchain uses since we use their text splitter for LLM2?

https://python.langchain.com/api_reference/_modules/langchain_text_splitters/character.html#CharacterTextSplitter.split_text

lib/TaskProcessing/SummaryProvider.php
edward-ly: In the loop, whatever summary output we get becomes the new prompt (input) in the next iteration. Regardless of whether chunking is disabled or not, this is the most reliable terminating condition we have.

lib/TaskProcessing/SummaryProvider.php
edward-ly: Loop is now rewritten to split on `newlines -> sentences -> words -> characters`. The only thing left that's bugging me at the moment is that when I test the assistant with longer text, the splitting doesn't actually work. I tested the function in the command-line and it works there, so maybe there's something different about the Nextcloud PHP environment that I'm missing.

lib/TaskProcessing/SummaryProvider.php
```diff
@@ -101,34 +101,108 @@ public function process(?string $userId, array $input, callable $reportProgress)
 			throw new RuntimeException('Invalid prompt');
 		}
 		$prompt = $input['input'];
-		$prompt = 'Summarize the following text. Detect the language of the text. Use the same language as the text.  Output only the summary.  Here is the text:' . "\n\n" . $prompt . "\n\n" . 'Here is your summary in the same language as the text:';
```
marcelklehr, line 104: We still need the instructions for the prompt, no?

marcelklehr:
Looks good and works in my tests. It seems we lost multilingual support though. When summarizing a German text that is longer than the chunk size the resulting summary was in english in my test.

edward-ly:
> Looks good and works in my tests. It seems we lost multilingual support though. When summarizing a German text that is longer than the chunk size the resulting summary was in english in my test.

In my testing, summarizing English text produced output in either Spanish or Italian, hence the change in prompt. If you want me to revert that change for now, we can do that.

marcelklehr:
>  In my testing, summarizing English text produced output in either Spanish or Italian

Oh, that's not intended of course :/ Ideally we should find a prompt that works for all languages to produce the summary in the same language as the input.

edward-ly:
Changed the prompt again after some basic prompt engineering and experimenting. Hopefully, this will produce more reliable results.

lib/TaskProcessing/SummaryProvider.php
```diff
@@ -122,11 +122,17 @@ public function process(?string $userId, array $input, callable $reportProgress)
 			try {
 				$completions = [];
 				if ($this->openAiAPIService->isUsingOpenAi() || $this->openAiSettingsService->getChatEndpointEnabled()) {
+					$summarySystemPrompt = 'Summarize the following text in a single paragraph and in the same language as the text.';
```
marcelklehr, line 124: Sorry i have to nitpick again, depending on the length of the input a single paragraph may be too short to include the relevant information :/

lib/TaskProcessing/SummaryProvider.php
```diff
@@ -101,34 +101,108 @@ public function process(?string $userId, array $input, callable $reportProgress)
...
+		$prompts = preg_split($splitByWord, $prompt, -1, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE);
+
+		if ($allChunksValid($prompts, $maxChars)) {
+			return $prompts;
 		}
 
-		throw new RuntimeException('No result in OpenAI/LocalAI response.');
+		// Split by number of characters in maximum chunk size
+		$splitBySize = sprintf('/(.{,%d})/su', $maxChars);
+		return preg_split($splitBySize, $prompt, -1, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE);
```
marcelklehr, line 206: For some reasons these preg_split calls all return only a single chunk for me having set chunk size to 1000 :thinking:

lib/TaskProcessing/SummaryProvider.php
edward-ly: If we put a cap on the length via the chunk size, then this shouldn't be too much of a problem; also, the advice from OpenAI says that the prompt should be as specific as possible. But even after some testing without the `in a single paragraph and`, the output length didn't change too much for me either, so maybe it's okay to leave it out for now.

lib/TaskProcessing/SummaryProvider.php
edward-ly: See my comment above:

> Apparently, the discrepancy is due to an outdated version of the PCRE library where having pcre.jit enabled changes the result, and the version bundled in nextcloud-docker-dev by default somehow isn't high enough. I'm not sure if there's an easy way to update the library, but after that update, the splitting should work again.

marcelklehr: I wonder if we should try using a different method then. I don't know how widespread these circumstances are. I think we can't require people to update pcre.

edward-ly: Maybe just using the same regexes but without the capturing groups can work too, I'll need some time to test them again either way.

edward-ly: Just pushed a new chunking algorithm which is similar to the previous one and still runs into the same library problem, but is much cleaner when it works. Manually splitting without regexes, especially by sentence, would be a chore and take more time and lines of code to implement. And if the regexes fail, it would just split the string every *n* characters as a fallback.

marcelklehr (approved):

marcelklehr:
Thanks for these adjustments, works well for me now!

julien-nc (approved):

edward-ly:
Rebased and adjusted after #167.

marcelklehr:
Woop woop :tada: 
Thanks @edward-ly !

PR #163, opened 2024-12-05
Feat: add change tone provider

janepie:


lib/TaskProcessing/ChangeToneProvider.php
```diff
@@ -0,0 +1,146 @@
...
+		return [
+			'tone_input' => 'friendler',
+		];
+	}
+
+	public function getOptionalInputShape(): array {
+		return [
+			'max_tokens' => new ShapeDescriptor(
+				$this->l->t('Maximum output words'),
+				$this->l->t('The maximum number of words/tokens that can be generated in the completion.'),
```
kyteinsky, line 69: tokens and words are different things. Generally for english, 1 token = 3/4 word. See the link for details: https://gptforwork.com/guides/openai-gpt3-tokens

let's use the term "tokens" exclusively here.

kyteinsky:
looks good, and works great otherwise!

lib/TaskProcessing/ChangeToneProvider.php
janepie: I can totally do that, just be aware that this is the same in every single provider right now. Maybe we should stay consistent?

kyteinsky: ah damn, calling for expert opinion on this @julien-nc

julien-nc: The users most likely don't know about the concept of token. Since tokens are approximately words, we could say "approximate number of words". Wdyt?

julien-nc (approved):
Works like a charm :grin: 
image

kyteinsky:
> The users most likely don't know about the concept of token. Since tokens are approximately words, we could say "approximate number of words". Wdyt?

yeah it's fine either way then. I forgot the fact that this text would be user facing, not just for admins.

kyteinsky (approved):

PR #167, opened 2024-12-10
Add ChatWithTools provider

julien-nc:
This will be used by the Agency feature which uses the tooling capabilities of LLMs.

This provider can be tested in the assistant because it dynamically generates a form for the task type (we will hide this task type later).

One small issue: We enforce all the input params but we might wanna set the tool_message (The result of tool calls in the last interaction).
If we give a fake/useless value to the OpenAI API, the request fails because the tool_message does not come after a "tool call" message.
To be discussed.

To test this provider, just replace `$toolMessage = $input['tool_message'];` by `$toolMessage = null;` in the provider and pass a dummy tool_message value in the form.
As all input fields have to be defined, also add at least one message in the "chat history".
You can use that in the "tools" input:
``` json
[{
    "type":"function",
    "function":{
        "description":"get the weather in a city",
        "name":"get_city_weather"
    }
}]
```
So if you ask for the weather in the city, you should get a tool call as a response. If not, you should get a classic response.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Nextcloud - OpenAI
  *
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -99,6 +99,9 @@ class Application extends App implements IBootstrap {
 			$context->registerTaskProcessingProvider(ReformulateProvider::class);
 			$context->registerTaskProcessingTaskType(ChangeToneTaskType::class);
 			$context->registerTaskProcessingProvider(ChangeToneProvider::class);
+			if (class_exists('OCP\\TaskProcessing\\TaskTypes\\TextToTextChatWithTools')) {
+				$context->registerTaskProcessingProvider(\OCA\OpenAi\TaskProcessing\TextToTextChatWithToolsProvider::class);
+			}
 		}
 		if ($this->appConfig->getValueString(Application::APP_ID, 't2i_provider_enabled', '1') === '1') {
 			$context->registerTaskProcessingProvider(TextToImageProvider::class);
--- a/lib/OldProcessing/Translation/TranslationProvider.php
+++ b/lib/OldProcessing/Translation/TranslationProvider.php
@@ -91,6 +91,7 @@ class TranslationProvider implements ITranslationProvider, IDetectLanguageProvid
 		try {
 			if ($this->openAiAPIService->isUsingOpenAi() || $this->openAiSettingsService->getChatEndpointEnabled()) {
 				$completion = $this->openAiAPIService->createChatCompletion($this->userId, $adminModel, $prompt, null, null, 1, 100);
+				$completion = $completion['messages'];
 			} else {
 				$completion = $this->openAiAPIService->createCompletion($this->userId, $prompt, 1, $adminModel, 100);
 			}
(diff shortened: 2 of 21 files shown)


lib/AppInfo/Application.php
```diff
@@ -99,6 +100,9 @@ public function register(IRegistrationContext $context): void {
 			$context->registerTaskProcessingProvider(ReformulateProvider::class);
 			$context->registerTaskProcessingTaskType(ChangeToneTaskType::class);
 			$context->registerTaskProcessingProvider(ChangeToneProvider::class);
+			if (class_exists('OCP\\TaskProcessing\\TaskTypes\\TextToTextChatWithTools')) {
+				$context->registerTaskProcessingProvider(\OCA\OpenAi\TaskProcessing\TextToTextChatWithToolsProvider::class);
+			}
```
marcelklehr, line 105: Nice! This will keep backwardcompatibility with NC 30

lib/OldProcessing/Translation/TranslationProvider.php
```diff
@@ -137,6 +138,7 @@ public function translate(?string $fromLanguage, string $toLanguage, string $tex
 
 			if ($this->openAiAPIService->isUsingOpenAi() || $this->openAiSettingsService->getChatEndpointEnabled()) {
 				$completion = $this->openAiAPIService->createChatCompletion($this->userId, $adminModel, $prompt, null, null, 1, PHP_INT_MAX);
+				$completion = $completion['messages'];
```
marcelklehr, line 141: summaries will need adjustment too, after edward's PR, I think

lib/Service/OpenAiAPIService.php
```diff
@@ -434,10 +450,17 @@ public function createChatCompletion(
...
+			'tool_calls' => [],
+		];
 
 		foreach ($response['choices'] as $choice) {
-			$completions[] = $choice['message']['content'];
+			if ($choice['finish_reason'] === 'tool_calls') {
+				$completions['tool_calls'][] = json_encode($choice['message']['tool_calls']);
+			} else {
+				$completions['messages'][] = $choice['message']['content'];
+			}
```
marcelklehr, line 463: It could be that we have a message and a tool_call in the same choice, I think.

lib/TaskProcessing/TextToTextChatWithToolsProvider.php
```diff
@@ -0,0 +1,139 @@
...
+		}
+		$systemPrompt = $input['system_prompt'];
+
+		if (!isset($input['tool_message']) || !is_string($input['tool_message'])) {
+			throw new RuntimeException('Invalid tool_message');
+		}
+		// TODO find a solution to allow passing no tool message
+		// OpenAI is rejecting the request if this param is set when there was no tool call done before
+		// and we are requiring this task input param
+		$toolMessage = $input['tool_message'];
```
marcelklehr, line 101: We can check if the tool message is the empty string and pass null instead

lib/OldProcessing/Translation/TranslationProvider.php
julien-nc: Yes, let's see who merges first.

lib/TaskProcessing/TextToTextChatWithToolsProvider.php
marcelklehr: But we don't want to use the normal form anyway, right?

julien-nc: Indeed but we might face this limitation with other task types. Let's talk about it later.

lib/Service/OpenAiAPIService.php
marcelklehr: The thing is, I think, that a "choice" is one way the model can answer. If we split this up into tool_call choices and messages choices, it's not clear which belong together.

Currently the result is 

```
array{messages: string[], tool_calls: string[]}
```


I think we need something like 

```
array<{message: string, tool_calls: string[]}>
```

marcelklehr (approved):
:rocket:

PR #169, opened 2024-12-19
Feat: add proofread provider

janepie:


lib/AppInfo/Application.php
```diff
@@ -105,6 +106,10 @@ public function register(IRegistrationContext $context): void {
 			if (class_exists('OCP\\TaskProcessing\\TaskTypes\\TextToTextChatWithTools')) {
 				$context->registerTaskProcessingProvider(\OCA\OpenAi\TaskProcessing\TextToTextChatWithToolsProvider::class);
 			}
+			if (class_exists('OCP\\TaskProcessing\\TaskTypes\\TextToTextProofread')) {
+				$context->registerTaskProcessingProvider(ProofreadProvider::class);
+			} else {
+			}
```
marcelklehr, line 112: Beep boop, unnecessary else clause :)

lib/AppInfo/Application.php
```diff
@@ -105,6 +106,9 @@ public function register(IRegistrationContext $context): void {
 			if (class_exists('OCP\\TaskProcessing\\TaskTypes\\TextToTextChatWithTools')) {
 				$context->registerTaskProcessingProvider(\OCA\OpenAi\TaskProcessing\TextToTextChatWithToolsProvider::class);
 			}
+			if (class_exists('OCP\\TaskProcessing\\TaskTypes\\TextToTextProofread')) {
+				$context->registerTaskProcessingProvider(ProofreadProvider::class);
```
marcelklehr, line 110: I *think* you will need to put the full class path in this line, otherwise PHP will try to load the class when loading Application.php and then will try to load `OCP\TaskProcessing\TaskTypes\TextToTextProofread` which will fail and the whole app will fail. I'm not sure if this is still the case, but I think that was a problem I ran into a few times.

janepie: Done :)

julien-nc (approved):
Works fine!

I tried to change the system prompt to force the llm to write correction recommendations in the same language as the submitted text but i couldn't get it to work.

I hope you don't mind, I made a small change to use the system prompt param of `OpenAiAPIService::createChatCompletion`.

marcelklehr:
> I tried to change the system prompt to force the llm to write correction recommendations in the same language as the submitted text but i couldn't get it to work

Mh, that's not ideal, we should fix this at some point

PR #177, opened 2025-01-15
 Make image generation work with  base64 responses (e.g. IONOS)

janepie:


lib/TaskProcessing/TextToImageProvider.php
```diff
@@ -148,6 +157,12 @@ public function process(?string $userId, array $input, callable $reportProgress)
 				$imageResponse = $client->get($url, $requestOptions);
 				$output['images'][] = $imageResponse->getBody();
 			}
+			foreach ($b64s as $b64) {
+				if ($b64 !== null) {
```
kyteinsky, line 161: null is already filtered above, this list should be all strings

kyteinsky:
Let's add an option for choosing the response type in the admin settings in image gen section for api endpoints other than official openai. Or even wtih official openai.

marcelklehr:
I agree with @janepie, making it work automagically is superior to adding an option, I think.

janepie:
We could delete it and just always use the default?

kyteinsky:
yeah let's do that :rocket:

julien-nc (approved):

PR #179, opened 2025-01-17
Support chatting with o1 models

julien-nc:
Use 'user' role for system prompt when o1-* models are used

refs https://github.com/nextcloud/assistant/issues/175


marcelklehr (approved):

kyteinsky:
replacement seems to be 'developer': https://community.openai.com/t/o1-models-do-not-support-system-role-in-chat-completion/953880/12

julien-nc:
@kyteinsky Yep, saw that, but it is not yet working:
https://community.openai.com/t/o1-models-do-not-support-system-role-in-chat-completion/953880/13

kyteinsky:
should've read further :see_no_evil:

PR #184, opened 2025-01-28
Improve use_max_completion_tokens_param default

julien-nc:
OpenAI introduced a new `max_completion_tokens` param for the chat completion endpoint and deprecated the `max_tokens` one.

This change enables `use_max_completion_tokens_param` by default for OpenAI and disables it by default for non-OpenAI.


kyteinsky (approved):

kyteinsky:
maybe the description can be tweaked a bit

> and deprecated the `max_tokens` one.

and removed the `max_tokens` one.

PR #185, opened 2025-01-28
Add a warning if the assistant app is not installed

julien-nc:
image

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -10,6 +10,12 @@
 		</h2>
 		<div id="openai-content">
 			<div>
+				<NcNoteCard v-if="!state.assistant_enabled" type="warning">
+					{{ t('integration_openai', 'The Assistant app is not enabled. You need it to use the features provided by the OpenAI/LocalAI integration app.') }}
+					<a class="external" :href="appSettingsAssistantUrl" target="_blank">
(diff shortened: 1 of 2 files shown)


kyteinsky (approved):

src/components/AdminSettings.vue
```diff
@@ -10,6 +10,12 @@
 		</h2>
 		<div id="openai-content">
 			<div>
+				<NcNoteCard v-if="!state.assistant_enabled" type="warning">
+					{{ t('integration_openai', 'The Assistant app is not enabled. You need it to use the features provided by the OpenAI/LocalAI integration app.') }}
+					<a class="external" href="https://apps.nextcloud.com/apps/assistant" target="_blank">
```
marcelklehr, line 15: It might be nicer to link to the assistant app in the apps page in nextcloud instead of apps.nc.com

julien-nc: Done

janepie (approved):
Hmm it's not totally correct as it still provides a translation provider that is usable without Assistant, but I'm fine with not highlighting that

julien-nc:
@janepie True. Let's keep it simple. I think the vast majority of people who want to use integration_openai need more than translation. Also, translation is implemented independently of the assistant only in Mail AFAIK. So the assistant is still needed to translate outside of Mail.

PR #186, opened 2025-01-30
Add admin setting to toggle authentication in the image retrieval request

### LINKED ISSUE #183: openai api compatibilty not there when text2image is being used from the assistant

3.4.0

30.0.5

firefox mainly but happens in all browsers

### Describe the Bug

I just tested with openrouter.ai and see similar behavior as with litellm when it comes to image creation
No text2image is possible using any vision model and I'm getting this:

A TaskProcessing core:text2image task with id 633 failed with the following message: OpenAI/LocalAI's text to image generation failed with: API request error: Client error: POST https://openrouter.ai/api/v1/images/generations resulted in a 405 Method Not Allowed response

with litellm used as router / proxy,  in the most current version of the assistant, it's failing with a malformed header or authorization header missing:

A TaskProcessing core:text2image task with id 628 failed with the following message: OpenAI/LocalAI's text to image generation failed with: Client error: GET  https://oaidalleapiprodscus.blob.core.windows.net/private/org-rKV85dZ3nzrjJACAPLeO1q19/user-n4qFiJ116cZjf8wOGhcM9Hku/img-hubxx80OTV4LkFh6cvG0BfKB.png?st=2025-01-27T14%3A31%3A18Z&se=2025-01-27T16%3A31%3A18Z&sp=r&sv=2024-08-04&sr=b&rscd=inline&rsct=image/png&skoid=d505667d-d6c1-4a0a-bac7-5c84a87759f8&sktid=a48cca56-e6da-484e-a814-9c849652bcb3&skt=2025-01-27T00%3A20%3A43Z&ske=2025-01-28T00%3A20%3A43Z&sks=b&skv=2024-08-04&sig=3OKm5gD%2BpGIC%2BL7MuF6U0feVmPeeJsyYtCm306VxGwE%3D resulted in a 403 Server failed to authenticate the request. Make sure  the value of Authorization header is formed correctly including the  signature. response

Openrouter.ai is opeanai compatible on their api endpoint, as well as litellm

### Expected Behavior

create an image as requested

### To Reproduce

use any model router / proxy api text2image endpoint. It fails everywhere but in native openAI / LocalAI.
I assume the authorization header is not being sent if text2image is being used. Tested with a script including the auth header and I instantly got an image back

julien-nc:
closes #183 

In some cases, like when using a proxy like litellm (configured to use OpenAI), we need to authenticate to make the generation request (as usual) BUT once we get the result (that contains a list of URLs) the requests to download the images must not be authenticated (or it fails with `403 Server failed to authenticate the request...`).

Initially, for the image retrieval requests, we decided to not authenticate if OpenAI was the configured service and to authenticate for any other service.
This is now a bit more flexible:
* Default is the previous behaviour
* There is a "manual" option to override and choose whether those requests should be authenticated or not

### DIFF
--- a/lib/Service/OpenAiSettingsService.php
+++ b/lib/Service/OpenAiSettingsService.php
@@ -25,6 +25,7 @@ class OpenAiSettingsService {
 		'default_stt_model_id' => 'string',
 		'default_image_model_id' => 'string',
 		'default_image_size' => 'string',
+		'image_request_auth' => 'boolean',
 		'chunk_size' => 'integer',
 		'max_tokens' => 'integer',
 		'use_max_completion_tokens_param' => 'boolean',
@@ -267,6 +268,7 @@ class OpenAiSettingsService {
 			'default_stt_model_id' => $this->getAdminDefaultSttModelId(),
 			'default_image_model_id' => $this->getAdminDefaultImageModelId(),
 			'default_image_size' => $this->getAdminDefaultImageSize(),
+			'image_request_auth' => $this->getIsImageRetrievalAuthenticated(),
 			'chunk_size' => strval($this->getChunkSize()),
 			'max_tokens' => $this->getMaxTokens(),
 			'use_max_completion_tokens_param' => $this->getUseMaxCompletionTokensParam(),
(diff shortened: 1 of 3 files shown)


lib/Service/OpenAiSettingsService.php
```diff
@@ -321,6 +323,16 @@ public function getTranslationProviderEnabled(): bool {
 		return $this->appConfig->getValueString(Application::APP_ID, 'translation_provider_enabled', '1') === '1';
 	}
 
+	/**
+	 * @return bool
+	 */
+	public function getIsImageRetrievalAuthenticated(): bool {
+		$serviceUrl = $this->getServiceUrl();
+		$isUsingOpenAI = $serviceUrl === '' || $serviceUrl === Application::OPENAI_API_BASE_URL;
+		$default = $isUsingOpenAI ? '0' : '1';
```
kyteinsky, line 332: we can make a function out of it in this file too at some point

kyteinsky (approved):

Issue #114, opened 2024-07-29
rpic failed reading parameters from request with LocalAi & stablediffusion


a6datta:
### Which version of integration_openai are you using?

2.0.3

### Which version of Nextcloud are you using?

29.0.4

### Which browser are you using? In case you are using the phone App, specify the Android or iOS version and device please.

Chrome on windows

### Describe the Bug

When I try to generate an image through next cloud assistant and by selecting "Generate Image" on the popup, it returns 400. I have correctly setup my local host as text chat is working fine using local ai. 

On the server side the log shows: ERR Server error error="rpc error: code = Unavailable desc = error reading from server: EOF" ip=172.17.0.1 latency=2.72452591s method=POST status=500 url=/v1/images/generations. 

To test local ai setup I tested with 
curl http://localhost:8082/v1/images/generations     -H "Content-Type: application/json"     -H "Authorization: Bearer api_key"         -d '{
        "prompt": "A cute puppy",
        "size": "256x256"
      }'
With this command, the image was generated. 

### Expected Behavior

Http status code 200

### To Reproduce

Launch next cloud assistant by clicking on the icon on top bar. Select "Generate Image" on the popup. Monitor response on the console.

XueSheng-GIT:
Issue seems to be the `size` which defaults to 1024x1024 but localai only supports up to 512x512.
https://github.com/nextcloud/integration_openai/blob/bb75ff88c9aafc6868c07af8e9b7b72670144788/lib/AppInfo/Application.php#L39

It seems there's currently no setting to overwrite this default. Thus, you'll have to modify it directly in the code.

a6datta:
It worked. Thanks.

kyteinsky:
@XueSheng-GIT 
> Issue seems to be the size which defaults to 1024x1024 but localai only supports up to 512x512.

That depends on the model use you, see an example from local ai using 1024x1024: https://localai.io/features/image-generation/#text-to-image

The image size can be changed from "Assistant modal -> Generate image -> triple dot menu (near the send button) -> Show advanced options"

XueSheng-GIT:
@kyteinsky Thanks for your response. I'm on latest (stable) NC29.0.6 and can't find those advanced options (see screenshot). Are you using a beta version?
![Bildschirmfoto vom 2024-09-05 16-53-58](https://github.com/user-attachments/assets/e1828cc0-6e1c-42d5-be07-8ecf15218195)

[...]

PR #193, opened 2025-02-17
Raise the default image size for compatibility

Rello:
Due to compatibility reasons, the default size should be set to 1024

- dall-e 2 can work with 512

but
- dall-e 3 can only work > 1024
https://platform.openai.com/docs/guides/images#size-and-quality-options
- IONOS can only work >1024
https://api.ionos.com/docs/inference-openai/v1/#tag/OpenAI-Compatible-Endpoints/operation/openaiCompatImagesGenerationsPost

so to make it compatible with the most, it should be adopted


julien-nc (approved):

Rello:
@julien-nc 
I just now finished the description :-)

kyteinsky:
we did a reverse change sometime ago for localai, but ig it depends on the model used there, (https://github.com/nextcloud/integration_openai/issues/114)
and since it completely breaks dall-e 3, let's go with this.

PR #197, opened 2025-02-28
feat: add support for together ai

### LINKED ISSUE #195: api keys from together AI seem not supported

latest

31.0.0

Firefox

### Describe the Bug

together.ai offers an openai compatible inference api endpoint offering 200+ models. To test this, subscribtion is supereasy. When trying to use https://api.together.xyz/v1 with a valid api key, it fails to retrive the model list from https://api.together.xyz/v1/models. The same works using curl (tested with:
curl -X GET "https://api.together.xyz/v1/models" -H "Authorization: Bearer tgp_v1_<my_api_key_unexposed>" -H "Content-Type: application/json".)
That it's the key which causes the trouble within nextclouds openai integration is an assumption since there are other reports about the new keys in openAI not working and because it works without issues using CURL.

### Expected Behavior

should work just like openAI or LocalAI

### To Reproduce

see description

kyteinsky:
fixes #195

### DIFF
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -77,7 +77,7 @@ class OpenAiAPIService {
 	 * @return boolean
 	 */
 	private function isModelListValid($models): bool {
-		if (!is_array($models)) {
+		if (!is_array($models) || !array_is_list($models)) {
 			return false;
 		}
 		if (count($models) === 0) {
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -111,10 +111,15 @@ class OpenAiAPIService {
 			$this->areCredsValid = false;
 			throw $e;
 		}
-		if (!isset($modelsResponse['data'])) {
+		if (isset($modelsResponse['error'])) {
 			$this->logger->warning('Error retrieving models: ' . \json_encode($modelsResponse));
 			$this->areCredsValid = false;
-			throw new Exception($this->l10n->t('Unknown models error'), Http::STATUS_INTERNAL_SERVER_ERROR);
+			throw new Exception($modelsResponse['error'], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+
+		if (!isset($modelsResponse['data'])) {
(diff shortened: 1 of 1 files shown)


lib/Service/OpenAiAPIService.php
```diff
@@ -111,10 +111,15 @@ public function getModels(string $userId): array {
...
 		}
-		if (!isset($modelsResponse['data'])) {
+		if (isset($modelsResponse['error'])) {
 			$this->logger->warning('Error retrieving models: ' . \json_encode($modelsResponse));
 			$this->areCredsValid = false;
-			throw new Exception($this->l10n->t('Unknown models error'), Http::STATUS_INTERNAL_SERVER_ERROR);
+			throw new Exception($modelsResponse['error'], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+
+		if (!isset($modelsResponse['data'])) {
```
julien-nc, line 120: Isn't it a bit optimistic to say
* the request didn't fail
* there is no data prop
* so the payload is the model list

?

In this case, I think we should try to check a little bit more if the response is valid. We could check if it's a list, take the first item and check it's valid. Wdyt?

kyteinsky: yeah it kind of depended on the api to report errors properly
sounds good

lib/Service/OpenAiAPIService.php
```diff
@@ -72,6 +72,26 @@ public function getServiceName(): string {
...
 	}
 
+	/**
+	 * @param mixed $models
+	 * @return boolean
+	 */
+	private function isModelListValid($models): bool {
+		if (!is_array($models)) {
+			return false;
+		}
```
julien-nc, line 82:
```suggestion
		if (!is_array($models) || !array_is_list($models)) {
			return false;
		}
```
Wdyt?

julien-nc (changes requested):

lib/Service/OpenAiAPIService.php
kyteinsky: ah didn't know about that

julien-nc (approved):

PR #204, opened 2025-03-19
ffmpeg video recording files as ogg

andreydruz:
ffmpeg video recording files as ogg in order to deal with 25mb limit of OpenAI

https://community.openai.com/t/whisper-api-increase-file-limit-25-mb/566754

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,8 @@
 	],
 	"require": {
 		"php": "^8.1 || ^8.2 || ^8.3",
-		"bamarni/composer-bin-plugin": "^1.8"
+		"bamarni/composer-bin-plugin": "^1.8",
+		"php-ffmpeg/php-ffmpeg": "^1.1"
 	},
 	"scripts": {
 		"lint": "find . -name \\*.php -not -path './vendor*' -print0 | xargs -0 -n1 php -l",
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -603,9 +603,10 @@ class OpenAiAPIService {
 		string $model = Application::DEFAULT_MODEL_ID,
 	): string {
 
+		$inputFilePath = '/mnt/ncdata' . $file->getPath();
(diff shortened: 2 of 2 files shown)


edward-ly (changes requested):
Apologies for not getting back to you earlier, but thank you for your contribution to the app! We agree that converting to opus/ogg would be beneficial too, However, using the `exec` function and shell commands is a big security risk. Not to mention, it is unclear what exactly the ffmpeg command does without any supporting comments (which are also prone to errors if not maintained). Please refactor the command by using the [`php-ffmpeg`](https://github.com/PHP-FFMpeg/PHP-FFMpeg) package instead.

A few other things we'd also like you to add:
1. In `FFMpeg\FFMpeg::create()`, do not automatically detect the ffmpeg binary path, and instead explicitly set the path from the PATH variable. If you want to, you can try to add a config setting that lets users define a custom path (with the PATH value being the default), but that is entirely optional.
2. Also optional, but if you can: please add a report/notification back to the UI saying how much audio was passed down to the API. That'll easily let us verify whether the conversion actually works; other optimizations like chunking or silence detection can come later.

If you have any additional questions, we'd be happy to help or provide additional resources.

DaphneMuller:
hello @andreydruz , do you think you would have the time and energy to follow up on the feedback that was provided earlier or shall we try to schedule some time to finish up the pull request? Let us know what you prefer :)

andreydruz:
Hi @edward-ly and @DaphneMuller,

About "php-ffmpeg package", we will work on this doing this month and will send updated MR.

About "few other things", currently we are successfully working without this and will not do this.

DaphneMuller:
hi! what's the reason you closed it? shall I reopen the issue so someone of my team can look at it upcoming week? thanks for your contribution!

andreydruz:
Hi @DaphneMuller,

The solution with shell command was tested on real servers for a number of weeks before I sent it.
Solution with PHP-FFmpeg package I can't test.
So I closed the MR.

We will continue to work with shell command on our server and once developer will have time to set up dev env, test it on dev server after it test it on prod server for a week or so - I will reopen MR.

PR #207, opened 2025-04-02
fix: do not add /v1 twice to the service url

kyteinsky:


lib/Migration/Version030104Date20241011192830.php
```diff
@@ -28,7 +28,7 @@ public function __construct(
 	 */
 	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
 		$value = $this->openAiSettingsService->getServiceUrl();
-		if ($value !== '') {
+		if ($value !== '' && strpos($value, '/v1') === false) {
```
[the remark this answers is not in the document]
kyteinsky: nice, thanks!

it's an old one that somehow seems to have run more than once
see the linked issue.

lib/Migration/Version030104Date20241011192830.php
```diff
@@ -28,7 +28,7 @@ public function __construct(
 	 */
 	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
 		$value = $this->openAiSettingsService->getServiceUrl();
-		if ($value !== '') {
+		if ($value !== '' && !str_ends_with($value, '/v1')) {
```
julien-nc, line 31: And what if someone has set `https://blabla/v1/`?

kyteinsky: ah I missed that `/` in a hurry :)

julien-nc (approved):

PR #210, opened 2025-04-30
feat(Text2Speech): Add support for text to speech

lukasdotcom:
This creates an implementation for text to speech. (https://github.com/nextcloud/server/pull/52051)

I don't really know for sure what to do about the voices because I couldn't find an endpoint that lists the voices. Right now its just hardcoded to one of openai's voices. Some ideas that I have are:
1. Make the input field text for the voice and just fail when an invalid voice is given
2. Detect when open ai endpoint is being used and then use hardcoded list
3. Let admin create a list of valid values to use

Openai's documentation for text to speech: https://platform.openai.com/docs/guides/text-to-speech#voice-options

### DIFF
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -931,7 +931,16 @@ class OpenAiAPIService {
 			if ($respCode >= 400) {
 				return ['error' => $this->l10n->t('Bad credentials')];
 			} else {
-				return json_decode($body, true) ?: [$body];
+				if ($response->getHeader('Content-Type') === 'application/json') {
+					$parsedBody = json_decode($body, true);
+					if ($parsedBody === null) {
+						$this->logger->warning('Could not JSON parse the response', ['body' => $body]);
+						return ['error' => 'Could not JSON parse the response'];
+					}
+					return $parsedBody;
+				} else {
+					return ['body' => $body];
+				}
 			}
 		} catch (ClientException|ServerException $e) {
 			$responseBody = $e->getResponse()->getBody();
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -751,12 +751,14 @@ class OpenAiAPIService {
 	 * @param string $prompt
 	 * @param string $model
 	 * @param string $voice
+	 * @param int $speed
 	 * @return array
 	 * @throws Exception
 	 */
 	public function requestSpeechCreation(
-		?string $userId, string $prompt, string $model, string $voice,
+		?string $userId, string $prompt, string $model, string $voice, int $speed = 1,
 	): array {
+		$this->logger->error('speed is ' . $speed);
 		if ($this->isQuotaExceeded($userId, Application::QUOTA_TYPE_SPEECH)) {
 			throw new Exception($this->l10n->t('Speech generation quota exceeded'), Http::STATUS_TOO_MANY_REQUESTS);
 		}
@@ -765,7 +767,8 @@ class OpenAiAPIService {
 			'input' => $prompt,
 			'voice' => $voice === Application::DEFAULT_MODEL_ID ? Application::DEFAULT_SPEECH_VOICE : $voice,
 			'model' => $model === Application::DEFAULT_MODEL_ID ? Application::DEFAULT_SPEECH_MODEL_ID : $model,
-			'response_format' => 'wav'
+			'response_format' => 'wav',
+			'speed' => $speed,
 		];
 
 		$apiResponse = $this->request($userId, 'audio/speech', $params, 'POST');
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -248,9 +248,9 @@ class OpenAiAPIService {
 				return $this->l10n->t('Image generation');
 			case Application::QUOTA_TYPE_TRANSCRIPTION:
 				return $this->l10n->t('Audio transcription');
-            case Application::QUOTA_TYPE_SPEECH:
-                return $this->l10n->t('Text to speech');
-            default:
+			case Application::QUOTA_TYPE_SPEECH:
+				return $this->l10n->t('Text to speech');
+			default:
 				return $this->l10n->t('Unknown');
 		}
 	}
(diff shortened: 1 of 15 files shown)


src/components/AdminSettings.vue
```diff
@@ -643,9 +691,16 @@ export default {
 						|| this.models[1]
 						|| this.models[0]
 
+					const defaultTtsModelId = this.state.default_stt_model_id || response.data?.default_stt_model_id
```
julien-nc, line 694: I know it's a draft PR. Just in case you are wondering:
```suggestion
					const defaultTtsModelId = this.state.default_speech_model_id || response.data?.default_stt_model_id
```
and you need to adjust the `onModelSelected` method.
In the backend, the `default_stt_model_id` needs to be declared as a setting key in `lib/Service/OpenAiSettingsService.php` (add it in `ADMIN_CONFIG_TYPES` and implement getter+setter).

julien-nc (changes requested):
Nicely done!
Can you use tabs for indentation in Php files?
You can fix all your files at once with `composer run cs:fix`

lib/TaskProcessing/TextToSpeechProvider.php
```diff
@@ -0,0 +1,146 @@
...
+
+
+		if (isset($input['voice']) && is_string($input['voice'])) {
+			$voice = $input['voice'];
+		} else {
+			$voice = $this->appConfig->getValueString(Application::APP_ID, 'default_speech_voice', Application::DEFAULT_MODEL_ID) ?: Application::DEFAULT_SPEECH_VOICE;
+		}
+
+		try {
+			$apiResponse = $this->openAiAPIService->requestSpeechCreation($userId, $prompt, $model, $voice);
```
kyteinsky, line 134: `speed` is not used here from `$input['speed']`

kyteinsky (changes requested):
thanks for the nice work!
It works wonderfully.

btw, the limit for the input is 4096 characters (https://platform.openai.com/docs/api-reference/audio/createSpeech#audio-createspeech-input). It would be a QoL improvement to split the sentences as such but it can be a feature for later.

lib/Service/OpenAiAPIService.php
```diff
@@ -894,7 +931,7 @@ public function request(?string $userId, string $endPoint, array $params = [], s
 			if ($respCode >= 400) {
 				return ['error' => $this->l10n->t('Bad credentials')];
 			} else {
-				return json_decode($body, true) ?: [];
+				return json_decode($body, true) ?: [$body];
```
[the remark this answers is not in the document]
lukasdotcom: I implemented a check for the response type and had it return the body, and also added error logging when the api returns a content type of json but json parsing fails.
https://github.com/lukasdotcom/integration_openai/blob/b565914dea5cfda4a83453ba2de7ffe2eff7a796/lib/Service/OpenAiAPIService.php#L942

lib/TaskProcessing/TextToSpeechProvider.php
lukasdotcom: Turns out speed doesn't exist as a parameter for the api. It had existed in the local tts that I was using so I removed all the references to speed.

lib/Service/OpenAiAPIService.php
```diff
@@ -742,6 +746,39 @@ public function getImageRequestOptions(?string $userId): array {
...
+	): array {
+		if ($this->isQuotaExceeded($userId, Application::QUOTA_TYPE_SPEECH)) {
+			throw new Exception($this->l10n->t('Speech generation quota exceeded'), Http::STATUS_TOO_MANY_REQUESTS);
+		}
+
+		$params = [
+			'input' => $prompt,
+			'voice' => $voice === Application::DEFAULT_MODEL_ID ? Application::DEFAULT_SPEECH_VOICE : $voice,
+			'model' => $model === Application::DEFAULT_MODEL_ID ? Application::DEFAULT_SPEECH_MODEL_ID : $model,
+			'response_format' => 'wav'
```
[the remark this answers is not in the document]
lukasdotcom: I've kept it for now as a constant because for image generation there isn't currently anyway to choose the format, but I can still make it be an optional input enum if that is wanted.

lib/TaskProcessing/TextToSpeechProvider.php
kyteinsky: speed does exist though
https://platform.openai.com/docs/api-reference/audio/createSpeech

lukasdotcom: I added back speed. Right now the selector for the speed has options that are kind of weird for a speed selector, and 5 is technically not valid for the official specifications.
image

kyteinsky: doesn't look like it can be helped. The `SmallNumberField` component is used instead of `NumberField` automatically when the default value is <10 https://github.com/nextcloud/assistant/blob/271d5fcde63231289ed60d2d9e0fb2210f9c0019/src/components/fields/TaskTypeField.vue#L88-L93

it would be nice to add a hint in the field's name that increasing the value decreases speed.
for the `5` value I'm not too sure about the best approach here, we can definitely return an error for >4 values in case of official openai as the provider but that error is not visible to the user.
maybe @julien-nc has any ideas?

kyteinsky: discussed with Julien

1. the label can be dynamically adjusted to add the hint `Valid values 1 to 4` if openai is used
2. if the value received in the provider is still out of bounds, we adjust it to lower and upper bounds, eg. if the value is 10, we make it 4 for openai and pass it along. No adjustment may be required for custom APIs since they may accept other values.

wdyt @lukasdotcom, does it address all the cases?

lukasdotcom: The valid values for openai are 0.25 to 4, but other than that everything else sounds good.

kyteinsky (commented):
some minor changes

lib/TaskProcessing/TextToSpeechProvider.php
kyteinsky: nice catch, and thanks for the fix!

src/components/AdminSettings.vue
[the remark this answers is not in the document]
lukasdotcom: Should be everything for Vue 3

kyteinsky (approved):
thanks a lot :rocket:

Issue #170, opened 2024-12-21
Implement task type: generate emoji


marcelklehr:
### Describe the feature you'd like to request

https://github.com/nextcloud/server/blob/master/lib/public/TaskProcessing/TaskTypes/GenerateEmoji.php

### Describe the solution you'd like

https://github.com/nextcloud/server/blob/master/lib/public/TaskProcessing/TaskTypes/GenerateEmoji.php

### Describe alternatives you've considered

-

PR #213, opened 2025-05-06
feat(Emoji): Implemnt emoji generation task

Lukasdotcom:
I tried it on a few prompts and a few different models (llama, gemma, chatgpt)  where it generated reasonable outputs.

Implements: #170


kyteinsky (approved):
thanks a lot!

PR #214, opened 2025-05-20
fix(TaskProcessing): update summary prompt

### LINKED ISSUE #209: Summarize always outputs spanish for gemma3 models
### Which version of assistant are you using?

2.4.0

30.0.10

Firefox, Chromium, etc.

### Describe the Bug

- Output is always in Spanish.

### Expected Behavior

- Output in the same language as input.

### To Reproduce

- Set up gemma3:12b via LocalAI Integration via Ollama.
- Use the Summarize tab on any arbitrary text.
- Output is always in Spanish (occasionally 1 in 10 times in Italian).

**Question:**
Is there any way to view or alter the **Chat User Instructions** for the Summary tool like there is for Chat Completions and Title Generation (at /settings/admin/ai)? 

It would also be useful for the **Translate** tool as well, since sometimes the model will give a bullet point line-by-line translation with the original text interleaved rather than just the results text. Giving a suitable pre-prompt in the Chat With AI or Generate Text tabs (such as "just output the translation, nothing else") allows for this trivially.

edward-ly:
Resolves #209 (or at least, that is the goal). A variation of this prompt has been tested to work on Gemma 3 12B along with several of the GPT models, but that is admittedly still not exhaustive by any means.

### DIFF
--- a/lib/TaskProcessing/SummaryProvider.php
+++ b/lib/TaskProcessing/SummaryProvider.php
@@ -126,7 +126,8 @@ class SummaryProvider implements ISynchronousProvider {
 			try {
 				$completions = [];
 				if ($this->openAiAPIService->isUsingOpenAi() || $this->openAiSettingsService->getChatEndpointEnabled()) {
-					$summarySystemPrompt = 'Summarize the following text in the same language as the text.';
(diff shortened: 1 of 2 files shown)


kyteinsky (approved):
nice!

PR #220, opened 2025-06-11
Migrate to Vue 3 and @nextcloud/vue 9

julien-nc:
RCs are available for `@nextcloud/vue` v9 that uses Vue 3.
There are many things to adjust. I think I covered everything.


kyteinsky (approved):
:rocket:

PR #226, opened 2025-07-01
Feat: Add asking question about picture

lukasdotcom:
The name for this in my opinion is not great. Some ideas I had: Image Question (Current), or Picture Chat. This should probably also be added as a task type on server.

### DIFF
--- a/lib/TaskProcessing/ImageQuestionProvider.php
+++ b/lib/TaskProcessing/ImageQuestionProvider.php
@@ -1,0 +1,159 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\OpenAi\TaskProcessing;
+
+use OCA\OpenAi\AppInfo\Application;
+use OCA\OpenAi\Service\OpenAiAPIService;
+use OCA\OpenAi\Service\OpenAiSettingsService;
+use OCP\Files\File;
+use OCP\IAppConfig;
+use OCP\IL10N;
+use OCP\TaskProcessing\EShapeType;
+use OCP\TaskProcessing\ISynchronousProvider;
+use OCP\TaskProcessing\ShapeDescriptor;
+use Psr\Log\LoggerInterface;
+use RuntimeException;
+
+class ImageQuestionProvider implements ISynchronousProvider {
+
+	public function __construct(
+		private OpenAiAPIService $openAiAPIService,
+		private OpenAiSettingsService $openAiSettingsService,
+		private IL10N $l,
+		private LoggerInterface $logger,
+		private IAppConfig $appConfig,
+		private ?string $userId,
+	) {
+	}
+
+	public function getId(): string {
+		return Application::APP_ID . '-image_question';
+	}
+
+	public function getName(): string {
+		return $this->openAiAPIService->getServiceName();
+	}
+
+	public function getTaskTypeId(): string {
+		return ImageQuestionTaskType::ID;
+	}
+
+	public function getExpectedRuntime(): int {
+		return $this->openAiAPIService->getExpTextProcessingTime();
+	}
+
+	public function getInputShapeEnumValues(): array {
+		return [];
+	}
+
+	public function getInputShapeDefaults(): array {
+		return [];
+	}
+
+
+	public function getOptionalInputShape(): array {
+		return [
+			'max_tokens' => new ShapeDescriptor(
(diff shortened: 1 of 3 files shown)


lib/TaskProcessing/ImageQuestionProvider.php
```diff
@@ -0,0 +1,159 @@
...
+	public function getOptionalInputShape(): array {
+		return [
+			'max_tokens' => new ShapeDescriptor(
+				$this->l->t('Maximum output words'),
+				$this->l->t('The maximum number of words/tokens that can be generated in the completion.'),
+				EShapeType::Number
+			),
+			'model' => new ShapeDescriptor(
+				$this->l->t('Model'),
+				$this->l->t('The model used to generate the completion'),
```
kyteinsky, line 70:
```suggestion
				$this->l->t('The maximum number of words/tokens that can be generated in the output.'),
				EShapeType::Number
			),
			'model' => new ShapeDescriptor(
				$this->l->t('Model'),
				$this->l->t('The model used to generate the output'),
```

lib/TaskProcessing/ImageQuestionProvider.php
```diff
@@ -0,0 +1,159 @@
...
+
+		if (!isset($input['image']) || !$input['image'] instanceof File || !$input['image']->isReadable()) {
+			throw new RuntimeException('Invalid input file');
+		}
+
+		$inputFile = base64_encode(stream_get_contents($input['image']->fopen('rb')));
+		$fileType = $input['image']->getMimeType();
+		if (!str_starts_with($fileType, 'image/')) {
+			throw new RuntimeException('Invalid input file type ' . $fileType);
+		}
```
kyteinsky, line 118: for openai, we can filter the mimetype here based on the supported ones: https://platform.openai.com/docs/guides/images-vision?api-mode=responses&format=url#image-input-requirements

lib/TaskProcessing/ImageQuestionProvider.php
```diff
@@ -0,0 +1,159 @@
...
+		try {
+			$systemPrompt = 'Take the users question and answer it based on the provided image. Ensure that the answer matches the language of the user\'s text input.';
+			$completion = $this->openAiAPIService->createChatCompletion($userId, $model, $prompt, $systemPrompt, [json_encode([
+				'role' => 'user',
+				'content' => [[
+					'type' => 'image_url',
+					'image_url' => [
+						'url' => 'data:' . $fileType . ';base64,' . $inputFile
+					]]
+				]
```
kyteinsky, line 145: formatting of the brackets can be improved a little

kyteinsky (commented):
Thanks!

lib/TaskProcessing/ImageQuestionProvider.php
kyteinsky: we let it run wild :horse:

lib/TaskProcessing/AnalyzeImageProvider.php
```diff
@@ -0,0 +1,174 @@
...
+	public function getId(): string {
+		return Application::APP_ID . '-analyze-image';
+	}
+
+	public function getName(): string {
+		return $this->openAiAPIService->getServiceName();
+	}
+
+	public function getTaskTypeId(): string {
+		return AnalyzeImageTaskType::ID;
```
kyteinsky, line 45: from the latest commit, task type id needs to be conditionally loaded from the server too

lukasdotcom: Fixed

julien-nc:
A simple approach would be to limit the number of images to a relatively low number like 20 or 30 and also limit sum of image sizes to something lower than 50 MB (or 40 MB seems reasonable to keep a margin for the prompt).
The task would fail if the input is above the limits.
The provider could set an informative message when failing, mentioning if it's because of the number of images or the total payload size.
Wdyt?

julien-nc:
I'm not in favor of sending a subset of the input images to make the request to the service succeed. There is no way for the user to know this has happened and why the result is no pertinent.

lukasdotcom:
> A simple approach would be to limit the number of images to a relatively low number like 20 or 30 and also limit sum of image sizes to something lower than 50 MB (or 40 MB seems reasonable to keep a margin for the prompt). The task would fail if the input is above the limits. The provider could set an informative message when failing, mentioning if it's because of the number of images or the total payload size. Wdyt?

I think failing is probably the better option definitely after seeing that some models can actually understand 100s of images.

I did just try out a few questions on about 400 pictures and google's models could answer the questions.
Eg: "There is a picture of a farmers market. What is the name of the market?" or "There is a picture of police officers on horses. How many horses are there?"
So surprisingly many files can be understood.

I'll probably just implement the official openai limits and fail if more than that is given.

PR #227, opened 2025-07-02
Implement audio chat provider

julien-nc:
Requires https://github.com/nextcloud/assistant/pull/291 (for the mp3 recording).
Related with https://github.com/nextcloud/server/pull/53759

If connected to OpenAI, we can use the chat completion endpoint with the `gpt-4o-audio-preview` model which accepts audio messages as input.

If not connected to OpenAI, it's safer to split the process in: STT -> LLM -> TTS.


lib/AppInfo/Application.php
```diff
@@ -126,6 +128,8 @@ public function register(IRegistrationContext $context): void {
 		if ($this->appConfig->getValueString(Application::APP_ID, 't2i_provider_enabled', '1') === '1') {
 			$context->registerTaskProcessingProvider(TextToImageProvider::class);
 		}
+		$context->registerTaskProcessingTaskType(AudioToAudioChatTaskType::class);
+		$context->registerTaskProcessingProvider(AudioToAudioChatProvider::class);
```
lukasdotcom, line 132: Something small I noticed while quickly looking through this is that you might want to only register this provider if they are using openai or have tts and stt enabled.

kyteinsky (commented):
just nitpicks :)

lib/TaskProcessing/AudioToAudioChatTaskType.php
```diff
@@ -0,0 +1,82 @@
...
+	public function __construct(
+		private IL10N $l,
+	) {
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getName(): string {
+		return $this->l->t('ioa Voice chat');
```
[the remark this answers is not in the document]
julien-nc: It was for debug purposes. Not there anymore.

lib/TaskProcessing/AudioToAudioChatProvider.php
```diff
@@ -0,0 +1,249 @@
...
+				throw new RuntimeException('OpenAI/LocalAI\'s text to speech generation failed: no speech returned');
+			}
+			return [
+				'output' => $apiResponse['body'],
+				'output_transcript' => $llmResult,
+				'input_transcript' => $inputTranscription,
+			];
+		} catch (\Exception $e) {
+			$this->logger->warning('OpenAI/LocalAI\'s text to speech generation failed with: ' . $e->getMessage(), ['exception' => $e]);
+			throw new RuntimeException('OpenAI/LocalAI\'s text to speech generation failed with: ' . $e->getMessage());
```
lukasdotcom, line 246:
```suggestion
			$this->logger->warning('LocalAI\'s text to speech generation failed with: ' . $e->getMessage(), ['exception' => $e]);
			throw new RuntimeException('LocalAI\'s text to speech generation failed with: ' . $e->getMessage());
```

lib/TaskProcessing/AudioToAudioChatProvider.php
```diff
@@ -0,0 +1,249 @@
...
+			return $result;
+		}
+
+		//////////////// 3 steps: STT -> LLM -> TTS
+		// speech to text
+		try {
+			$inputTranscription = $this->openAiAPIService->transcribeFile($userId, $inputFile, false, $sttModel);
+		} catch (Exception $e) {
+			$this->logger->warning('OpenAI\'s Whisper transcription failed with: ' . $e->getMessage(), ['exception' => $e]);
+			throw new RuntimeException('OpenAI\'s Whisper transcription failed with: ' . $e->getMessage());
```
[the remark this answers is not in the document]
lukasdotcom: Small nitpick with that this technically can't ever be openai

lib/TaskProcessing/AudioToAudioChatProvider.php
```diff
@@ -0,0 +1,249 @@
...
+
+		/////////////// Using the chat API if connected to OpenAI
+		if ($this->openAiAPIService->isUsingOpenAi()) {
+			$b64Audio = base64_encode($inputFile->getContent());
+			$extraParams = [
+				'modalities' => ['text', 'audio'],
+				'audio' => ['voice' => $outputVoice, 'format' => 'mp3'],
+			];
+			$completion = $this->openAiAPIService->createChatCompletion(
+				$userId, 'gpt-4o-audio-preview', null, $systemPrompt, $history, 1, 1000,
```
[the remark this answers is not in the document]
julien-nc: Yep, changed the llm default, it's now used in the one shot request to OpenAI's chat endpoint.

lib/TaskProcessing/AudioToAudioChatProvider.php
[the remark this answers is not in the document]
julien-nc: This only appears in logs anyway. I can change it to the app ID or the service name.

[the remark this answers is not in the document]
julien-nc: I went with the service name and it falls back to the app ID if no service name is defined.

lukasdotcom (approved):
LGTM

lib/TaskProcessing/AudioToAudioChatProvider.php
```diff
@@ -127,7 +127,13 @@ public function getOutputShapeEnumValues(): array {
...
 
 	public function getOptionalOutputShape(): array {
-		return [];
+		return [
+			'audio_id' => new ShapeDescriptor(
+				$this->l->t('Remote audio ID'),
+				$this->l->t('The ID of the audio response returned by the remote service'),
+				EShapeType::Text
+			),
+		];
```
lukasdotcom, line 136: I was looking at the documentation and I'm not really sure how this can be used. If I understand it correctly it only helps with getting the response, but we are already returning the response anyway under output/output_transcription (feel free to correct me here). For reference: https://platform.openai.com/docs/api-reference/chat/get Also I'm not sure how long this would be stored.

lukasdotcom (changes requested):
Splitting it into two functions is a good idea, and I have some feedback on the audio id portion.

lib/TaskProcessing/AudioToAudioChatProvider.php
julien-nc: I saw that https://community.openai.com/t/gpt-4o-audio-preview-responds-in-text-not-audio/1006486/4 and tried it and it works just fine. With the assistant, i build a history that is full of 
```
{"role":"assistant","audio":{"id":"audio_686e7a6c46ec81919a688020d53e7dbf"}
```
and the remote service (I could only test with OpenAI) gets it.

julien-nc: We never need to "get the message later", we don't use the audio ID like that. We use it in the chat history of the following requests.

lukasdotcom: Okay that makes sense. The documentation was not very helpful for this at all IMO. You can probably than ignore my second comment too.

lib/TaskProcessing/AudioToAudioChatProvider.php
```diff
@@ -0,0 +1,302 @@
...
+					$this->logger->warning($serviceName . ' text to speech generation failed: no speech returned');
+					throw new RuntimeException($serviceName . ' text to speech generation failed: no speech returned');
+				}
+				$output = $apiResponse['body'];
+			} catch (\Exception $e) {
+				$this->logger->warning($serviceName . ' text to speech generation failed with: ' . $e->getMessage(), ['exception' => $e]);
+				throw new RuntimeException($serviceName . ' text to speech generation failed with: ' . $e->getMessage());
+			}
+		} else {
+			$output = base64_decode($message['audio']['data']);
```
kyteinsky, line 239: this audio data expires at a certain time `expires_at`. Should we check that and fallback to the above TTS approach to get the new audio and audio id?

lukasdotcom (approved):
Other than what @kyteinsky said everything else LGTM.

PR #228, opened 2025-07-04
feat: use outline icons instead of solid ones

kyteinsky:
### In admin settings, the help icon and the authentication icons changed

image

image

### In personal settings, the key icon was changed, and the heading is left-aligned for consistency. The basic auth UI and logic is also ported from the admin settings.

image

image

For the vue3 migration, it looks like `NcInputField` and `NcTextField` need icon slots now.


lukasdotcom (approved):
LGTM 🚀

Issue #200, opened 2025-03-06
Inform user if the account is running out of tokens


julien-nc:
### Describe the feature you'd like to request

It would be nice to inform the user or the admin when the quota is exceeded on the service side.

I don't know if the error is generic when using the API of various services.
I also don't know if there is an endpoint to retrieve the quota information.

### Describe the solution you'd like

It could be:
* a toast message when the task fails
* a notification to the admin
* a warning in the admin settings

### Describe alternatives you've considered

Leave it like it is now. The error is written in the logs and in the `error_output` of the task db item.

PR #254, opened 2025-08-12
feat: add notification when quota exceeded

lukasdotcom:
Discussed to be a background job, but I think it would probably be better to notify when the quota is exceeded as this is already checked by default. This also means that the assistant error has more information when the quota is exceeded in a separate notification.
Resolves: #200

### DIFF
--- a/tests/unit/Quota/QuotaTest.php
+++ b/tests/unit/Quota/QuotaTest.php
@@ -13,7 +13,6 @@
 namespace OCA\OpenAi\Tests\Unit\Quota;
 
 use Exception;
-use OC;
 use OCA\OpenAi\AppInfo\Application;
 use OCA\OpenAi\Db\QuotaUsageMapper;
 use OCA\OpenAi\Service\OpenAiAPIService;
@@ -55,7 +54,7 @@ class QuotaTest extends TestCase {
 		parent::setUpBeforeClass();
 		$backend = new Dummy();
 		$backend->createUser(self::TEST_USER1, self::TEST_USER1);
-		OC::$server->get(IUserManager::class)->registerBackend($backend);
+		\OCP\Server::get(IUserManager::class)->registerBackend($backend);
 	}
 
 	protected function setUp(): void {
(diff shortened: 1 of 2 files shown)


kyteinsky (commented):
nice, this approach is indeed better.

tests/unit/Quota/QuotaTest.php
```diff
@@ -0,0 +1,124 @@
...
+	 * @var MockObject|ICacheFactory
+	 */
+	private $cacheFactory;
+	private QuotaUsageMapper $quotaUsageMapper;
+
+	public static function setUpBeforeClass(): void {
+		parent::setUpBeforeClass();
+		$backend = new Dummy();
+		$backend->createUser(self::TEST_USER1, self::TEST_USER1);
+		OC::$server->get(IUserManager::class)->registerBackend($backend);
```
kyteinsky, line 58: let's use the OCP API `\OCP\Server::get()`
in the other places as well.

lib/Service/OpenAiAPIService.php
```diff
@@ -252,8 +255,21 @@ public function isQuotaExceeded(?string $userId, int $type): bool {
...
+					->setUser($userId)
+					->setDateTime(new DateTime())
+					->setObject('quota_exceeded', (string)$type)
+					->setSubject('quota_exceeded', ['type' => $type]);
+				$this->notificationManager->notify($notification);
+				$cache->set('quota_exceeded_' . $userId . '_' . $type, true, 3600);
+			}
+			return true;
+		}
+		return false;
```
kyteinsky, line 272: [optional]
wdyt of converting `if ($quotaUsage >= $quota) {` to a negative check and returning early?
```php
if ($quotaUsage < $quota) {
    return false;
}
```

then the pattern becomes `return false`, `return false`, ..., `return true` which then becomes much easier to read and understand.

tests/unit/Quota/QuotaTest.php
lukasdotcom: Its just one more ctr+f 🤷

kyteinsky (approved):
looks good!

would you mind organizing your commits into two parts? The second one becuase it affects more than quota related changes.
"feat: add notification when quota exceeded" and "chore: use ocp server get"
or you can merge the whole commit as one using the squash button here in github.
the incremental changes in different commits is nice for PRs but when looking back, you see individual commits and comments in code.

Issue #250, opened 2025-08-06
Define quota period in dates and add usage report generation


kyteinsky:
### Describe the feature you'd like to request

1. Convert the quota period from time interval (last 30 days from today) to set dates which start at first day of the month and ends in 1 month, 2 months, etc.
2. Report generation on the usage of the users of the instance. Filters can be based on groups (see #149) and time period of the quotas.

### Describe the solution you'd like

/

### Describe alternatives you've considered

/

nimishavijay:
Subbing in for @marcoambrosini 

Not sure about the scope of this issue, but as I understand we want to replace the existing "Quota enforcement period" to a more custom time period. I have thought about this similar to the repeating event in a calendar, but feel free to scope down if this is too detailed:

- [ ] It would follow the structure
    ```
    Reset every [ number ] [ time period ]
    [ options to select date within time period ]
    [ summary ]
    ```
- [ ] As the time period changes we have to add different fields to select the exact date/day of reset 
  - [ ] Day: nothing required
  - [ ] Week: day of the week
    - [ ] To show the day of the week we can use a [NcCheckboxRadioSwitch ](https://nextcloud-vue-components.netlify.app/#/Components/NcCheckboxRadioSwitch)with the alternative button
    - [ ] That component would need some custom styling to make it look like the mockup (remove border, use --color-primary-element-light bg)
  - [ ] Month: specific date on the month, or specific day of the week in the month
    - [ ] For specific date, a grid would be best, but if that is out of scope a simple text field could also work
    - [ ] For specific day of the week (eg second Saturday), the options can be `first`, `second`, `third`, `fourth`, and `last`
  - [ ] Year: month + specific date on the month, or specific day of the week in the month
- [ ] The summary would immediately reflect the choice of the user. We can use a success NcNotecard for this
- [ ] What would be important is making sure that the summary string makes sense. It should always follow the format 
  `Quota will be reset every [number] [period] on [date/day]`
  There will probably be a few annoying edge cases like singular nouns when number is 1, etc. 

<img width="721" height="200" alt="Image" src="https://github.com/user-attachments/assets/62497ca1-c0e2-43b2-9bf4-8563cacdfd8e" />

<img width="720" height="227" alt="Image" src="https://github.com/user-attachments/assets/6cc859a7-a60e-4eab-8f01-c037c6ea7966" />

[...]

PR #256, opened 2025-08-15
feat: add monthly quota periods

lukasdotcom:
This does the minimum to fix this adding a monthly option for period that resets every month or exact multiple of months. This could also be used for yearly as long as a start date of January is ok. The date range was originally planned to be put into the admin settings, but I think it looks better in personal settings.
Resolves part of: #250
## Admin Settings
image
image

## Personal Settings
image
image

### DIFF
--- a/lib/Service/OpenAiSettingsService.php
+++ b/lib/Service/OpenAiSettingsService.php
@@ -65,6 +65,8 @@ class OpenAiSettingsService {
 	}
 
 	/**
+	 * Gets the timestamp of the beginning of the quota period
+	 *
 	 * @return int
 	 * @throws Exception
 	 */
@@ -93,6 +95,9 @@ class OpenAiSettingsService {
 	}
 
 	/**
+	 * Gets the timestamp of the end of the quota period
+	 * if the period is floating, then this will be the current time
+	 *
 	 * @return int
 	 * @throws Exception
 	 */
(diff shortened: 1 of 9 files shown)


lib/Service/OpenAiSettingsService.php
```diff
@@ -62,6 +64,62 @@ public function __construct(
 	) {
 	}
 
+	/**
+	 * @return int
+	 * @throws Exception
+	 */
```
julien-nc, line 70: Can you add a description here?

lib/Service/OpenAiSettingsService.php
```diff
@@ -62,6 +64,62 @@ public function __construct(
...
+				$periodStart = $periodStart->sub(new DateInterval('P' . $remainder . 'M'));
+			}
+		}
+		return $periodStart->getTimestamp();
+	}
+
+	/**
+	 * @return int
+	 * @throws Exception
+	 */
```
julien-nc, line 98: And here?

julien-nc (changes requested):
When there is no value for quota_period, one loads the admin settings page, switches `Reset quota every` from 30 days to 30 months, then it fails to save because it tries to save without the "day" param. It should be one by default if it's not set, just like it is visually 1 by default.

Could you also add a bit more descriptions of what the methods do?

lukasdotcom:
> When there is no value for quota_period, one loads the admin settings page, switches `Reset quota every` from 30 days to 30 months, then it fails to save because it tries to save without the "day" param. It should be one by default if it's not set, just like it is visually 1 by default.
> 
> Could you also add a bit more descriptions of what the methods do?

I remember that issue, but forgot about it when I manually fixed the config. Is fixed and rebased now.

Issue #262, opened 2025-08-26
i18n: Remove trailing slash and add comment for translators


rakekniven:
https://github.com/nextcloud/integration_openai/blob/b5841f38ebbe0181d4d3139ac71594ffdd381316/src/components/QuotaPeriodPicker.vue#L15

`Quota enforcement time ` -> `Quota enforcement time`

"Reset quota every" Please add a comment to source code for context

Issue #260, opened 2025-08-26
i18n: Context needed related to string about quota management


rakekniven:
https://github.com/nextcloud/integration_openai/blob/f8f81a8584fa3720789187baa945abcdfcb34edd/src/components/QuotaPeriodPicker.vue#L89

Question from Transifex:

What will be in the "day" variable? Is it a number (like 3) or word (like first) or something else?

@lukasdotcom Found you using git blame.

lukasdotcom:
It should be a number. Eg: `Quota will reset all users every month on day 10`

PR #265, opened 2025-08-27
i18n: context needed for the quota reset day

lukasdotcom:
Resolves: #260
Resolves: #262

### DIFF
--- a/src/components/QuotaPeriodPicker.vue
+++ b/src/components/QuotaPeriodPicker.vue
@@ -11,9 +11,7 @@
 		</NcNoteCard>
 		<div class="line">
 			<span class="text-cell">
-				{{
-					floating ? t('integration_openai', 'Quota enforcement time ') : t('integration_openai', 'Reset quota every')
-				}}
+				{{ textBeforeInput }}
 			</span>
 			<NcInputField
 				id="quota"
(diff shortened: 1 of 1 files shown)


src/components/QuotaPeriodPicker.vue
```diff
@@ -86,7 +84,12 @@ export default {
 		resetText() {
 			return this.floating
 				? n('integration_openai', 'Quota will be enforced based on last %n day of usage', 'Quota will be enforced based on last %n days of usage', this.value.length)
-				: n('integration_openai', 'Quota will reset all users every month on day {day}', 'Quota will reset for all users every %n months on day {day}', this.value.length, { day: this.value.day })
+				: n('integration_openai', 'Quota will reset all users every month on day {day}', 'Quota will reset for all users every %n months on day {day}', this.value.length, { day: this.value.day }) // TRANSLATORS: Day is a number like 1, 12, etc
```
kyteinsky, line 87:
```suggestion
				: n('integration_openai', 'Quota will reset for all the users every month on day no. {nth_day}', 'Quota will reset for all the users every %n months on day no. {nth_day}', this.value.length, { nth_day: this.value.day }) // TRANSLATORS: "nth_day" is a number like 1, 12, etc
```
what do you think?

lukasdotcom: I like that more, but did change no. to fully spelled out number. Not sure if using the shorthand would be a good idea.

kyteinsky (approved):
nice :rocket:

Issue #264, opened 2025-08-27
Allow to force language for STT


julien-nc:
### Describe the feature you'd like to request

STT sometimes does not write the output in the language that was spoken. It would be nice to optionally allow the users to force the STT language.

### Describe the solution you'd like

There is a `language` param for the `/audio/transcriptions` endpoint (see https://platform.openai.com/docs/api-reference/audio/createTranscription) to force the language detection.

PR #266, opened 2025-08-29
feat: specify language for AudioToText

lukasdotcom:
Resolves: #264 
I tested it with french on english audio and it did use french quotation marks `<<` instead of the english ones `"`.
image

### DIFF
--- a/lib/TaskProcessing/AudioToTextProvider.php
+++ b/lib/TaskProcessing/AudioToTextProvider.php
@@ -15,7 +15,6 @@ use OCA\OpenAi\Service\OpenAiAPIService;
 use OCP\Files\File;
 use OCP\IAppConfig;
 use OCP\IL10N;
-use OCP\L10N\IFactory;
 use OCP\TaskProcessing\EShapeType;
 use OCP\TaskProcessing\ISynchronousProvider;
 use OCP\TaskProcessing\ShapeDescriptor;
@@ -30,7 +29,6 @@ class AudioToTextProvider implements ISynchronousProvider {
 		private OpenAiAPIService $openAiAPIService,
 		private LoggerInterface $logger,
 		private IAppConfig $appConfig,
-		private IFactory $l10nFactory,
 		private IL10N $l,
 	) {
 	}
@@ -72,11 +70,12 @@ class AudioToTextProvider implements ISynchronousProvider {
 			return new ShapeEnumValue($language[1], $language[0]);
 		}, Application::AUDIO_TO_TEXT_LANGUAGES);
 		$detectLanguageEnumValue = new ShapeEnumValue($this->l->t('Detect language'), 'detect_language');
-		return ['language' => array_merge([$detectLanguageEnumValue], $languageEnumValues)];
+		$defaultLanguageEnumValue = new ShapeEnumValue($this->l->t('Default'), 'default');
+		return ['language' => array_merge([$detectLanguageEnumValue, $defaultLanguageEnumValue], $languageEnumValues)];
 	}
 
 	public function getOptionalInputShapeDefaults(): array {
-		return ['language' => 'detect_language'];
+		return ['language' => 'default'];
 	}
 
 	public function getOutputShapeEnumValues(): array {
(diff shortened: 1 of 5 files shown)


lib/TaskProcessing/AudioToTextProvider.php
```diff
@@ -53,15 +60,25 @@ public function getInputShapeDefaults(): array {
...
+			$this->l->t('Language'),
+			$this->l->t('The language of the audio file'),
+			EShapeType::Enum
+		)];
 	}
 
 	public function getOptionalInputShapeEnumValues(): array {
-		return [];
+		$coreL = $this->l10nFactory->getLanguages();
+		$languages = array_merge($coreL['commonLanguages'], $coreL['otherLanguages']);
```
kyteinsky, line 72: maybe we can hardcode the language list for openai
https://platform.openai.com/docs/guides/speech-to-text/supported-languages#supported-languages

lukasdotcom: Personally I think the hard coding should be done for all of them not just openai. The language list for openai is a lot larger than the one I had previously and is just whisper's anyway. The model a lot of other providers use. I did make the list use endonyms just like before (name of the language in its own language) so translations shouldn't be necessary.

kyteinsky: fair enough. The names of the languages in the NC selected language might still be good to have, can be discussed in this separate issue though: #269

lib/AppInfo/Application.php
```diff
@@ -79,6 +79,8 @@ class Application extends App implements IBootstrap {
...
...ارسی'], ['lv', 'Latviešu'], ['bn', 'বাংলা'], ['sr', 'Српски'], ['az', 'Azərbaycanca / آذربايجان'], ['sl', 'Slovenščina'], ['kn', 'ಕನ್ನಡ'], ['et', 'Eesti'], ['mk', 'Македонски'], ['br', 'Brezhoneg'], ['eu', 'Euskara'], ['is', 'Íslenska'], ['hy', 'Հայերեն'], ['ne', 'नेपाली'], ['mn', 'Монгол'], ['bs', 'Bosanski'], ['kk', 'Қазақша'], ['sq', 'Shqip'], ['sw', 'Kiswahili'], ['gl', 'Galego'], ['mr', 'मराठी'], ['pa', 'ਪੰਜਾਬੀ / पंजाबी / پنجابي'], ['si', 'සිංහල'], ['km', 'ភាសាខ្មែរ'], ['sn', 'chiShona'], ['yo', 'Yorùbá'], ['so', 'Soomaaliga'], ['af', 'Afrikaans'], ['oc', 'Occitan'], ['ka', 'ქართული'], ['be', 'Беларуская'], ['tg', 'Тоҷикӣ'], ['sd', 'सिनधि'], ['gu', 'ગુજરાતી'], ['am', 'አማርኛ'], ['yi', 'ייִדיש'], ['lo', 'ລາວ / Pha xa lao'], ['uz', 'Ўзбек'], ['fo', 'Føroyskt'], ['ht', 'Krèyol ayisyen'], ['ps', 'پښتو'], ['tk', 'Туркмен / تركمن'], ['nn', 'Norsk (nynorsk)'], ['mt', 'bil-Malti'], ['sa', 'संस्कृतम्'], ['lb', 'Lëtzebuergesch'], ['my', 'Myanmasa'], ['bo', 'བོད་ཡིག / Bod skad'], ['tl', 'Tagalog'], ['mg', 'Malagasy'], ['as', 'অসমীয়া'], ['tt', 'Tatarça'], ['haw', 'ʻŌlelo Hawaiʻi'], ['ln', 'Lingála'], ['ha', 'هَوُسَ'], ['ba', 'Башҡорт'], ['jw', 'ꦧꦱꦗꦮ'], ['su', 'Basa Sunda'], ['yue', '粤语']];
```
kyteinsky, line 82: nitp: `WHISPER_LANGUAGES`* but maybe can be named indicating the task type since later languages for translation will be added here too

kyteinsky (approved):
:rocket:

lukasdotcom:
Updated the pr to now be able to set a default in the user settings for the language that is used. Still supports the other way of selecting the language for a specific transcribe audio task.

PR #277, opened 2025-10-06
Store the model list in the DB

### LINKED ISSUE #272: App Causes Dashboard Hang Due to Blocking Network Request When LocalAI Server is Unreachable

3.7.1

31.0.9.1 (Nextcloud AIO)

Brave & Firefox

### Describe the Bug

The integration_openai app appears to perform a synchronous network request to fetch OpenAI models during the loading of the Nextcloud dashboard. If the configured LocalAI server is crashed or otherwise unreachable, this request times out after an excessively long period (observed as 160+ seconds), causing the entire web interface to hang post-authentication. This makes the dashboard and subsequent pages unresponsive until the timeout resolves or the app is disabled.

This behavior is problematic because it blocks the core Nextcloud UI functionality, even for users not actively using AI features.

Disabling the integration_openai app resolves the issue immediately, confirming it's the source of the hang.

Additional Context:

The issue appeared after the LocalAI server crashed; prior to that, everything functioned normally.

Desktop sync, API endpoints, and WebDAV continue to work fine during the hang, indicating the problem is isolated to the web UI/dashboard load.

This was first observed on September 26, 2025, with up-to-date Nextcloud AIO components and has taken me several days to isolate.

### Expected Behavior

The model fetch should ideally be asynchronous, fail gracefully with error handling, or be deferred until relevant action is initiated, rather than running on every dashboard load.

### To Reproduce

1. Install and enable the integration_openai app in Nextcloud.
2. Configure it to use a LocalAI server (e.g., via the app's settings with a custom base URL for LocalAI).
3. Ensure the LocalAI server is running initially and verify that models can be fetched successfully.
4. Crash or shut down the LocalAI server to make it unreachable.
5. Log into the Nextcloud web interface.
6. After successful authentication, attempt to load the dashboard (/apps/dashboard).
7.Observe the hang/delay of 160+ seconds before the page loads or times out.

julien-nc:
closes #272

This prevents the model list to be obtained with a network request to the external service when populating the enum values of the task processing providers (which is a problem when the remote service is unreachable).

We now store the model list in a permanent storage (the DB) and it is only refreshed:
* when the settings page is accessed
* on upgrade to 3.9.0
* once a day in a bg job

### DIFF
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -39,7 +39,6 @@ use function json_encode;
 class OpenAiAPIService {
 	private IClient $client;
 	private ?array $modelsMemoryCache = null;
-	private ?bool $areCredsValid = null;
 
 	public function __construct(
 		private LoggerInterface $logger,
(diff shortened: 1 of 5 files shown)


lib/Service/OpenAiAPIService.php
```diff
@@ -111,68 +110,77 @@ private function isModelListValid($models): bool {
...
+			// if we don't need to refresh to model list and it's not been found in the cache, it is obtained from the DB
+			$modelsObjectString = $this->appConfig->getValueString(Application::APP_ID, 'models', '{"data":[],"object":"list"}');
+			$fallbackModels = [
+				'data' => [],
+				'object' => 'list',
+			];
+			try {
+				$newCache = json_decode($modelsObjectString, true) ?? $fallbackModels;
+			} catch (Throwable $e) {
+				$newCache = $fallbackModels;
```
marcelklehr, line 162: Might be nice to have a warning in the logs for this case

lib/Service/OpenAiAPIService.php
```diff
@@ -111,68 +110,77 @@ private function isModelListValid($models): bool {
...
 			}
+			$cache->set($userId !== null ? $userCacheKey : $adminCacheKey, $newCache, Application::MODELS_CACHE_TTL);
+			$this->modelsMemoryCache = $newCache;
+			return $newCache;
 		}
 
+		// we know we are refreshing so we clear the caches and make the network request
+		$adminCacheKey = Application::MODELS_CACHE_KEY . '-main';
```
marcelklehr, line 170: Can we move these to the top of the method, so we have them only once?

lib/Service/OpenAiAPIService.php
```diff
@@ -111,68 +110,77 @@ private function isModelListValid($models): bool {
...
+			// if the user has an API key or uses basic auth, skip the admin cache
+			if (!(
+				$this->openAiSettingsService->getUserApiKey($userId, false) !== ''
+				|| (
+					$this->openAiSettingsService->getUseBasicAuth()
+					&& $this->openAiSettingsService->getUserBasicUser($userId) !== ''
+					&& $this->openAiSettingsService->getUserBasicPassword($userId) !== ''
+				)
+			)) {
+				// if no user cache or userId is null, try to get from the admin cache
```
marcelklehr, line 146: looks like this actually is not run if userId is null 🤔

julien-nc: Jup

marcelklehr (approved):

PR #295, opened 2025-11-18
feat(TextToImage, TextToSpeech): Add watermark and metadata

marcelklehr:
to announce that the files are created by AI

Examples:
[41-48.mp3](https://github.com/user-attachments/files/23603812/41-48.mp3)
[31-44.jpg](https://github.com/user-attachments/assets/457ec8d8-2a11-438c-b697-02f09858b998)

### DIFF
--- a/lib/TaskProcessing/TextToSpeechProvider.php
+++ b/lib/TaskProcessing/TextToSpeechProvider.php
@@ -124,7 +124,7 @@ class TextToSpeechProvider implements ISynchronousWatermarkingProvider {
 		$prompt = $input['input'];
 
 		if ($includeWatermark) {
-			$prompt .= "\n\n" . 'This was generated using Artificial Intelligence.';
+			$prompt .= "\n\n" . $this->l->t('This was generated using Artificial Intelligence.');
 		}
 
 		if (isset($input['model']) && is_string($input['model'])) {
--- a/tests/unit/Providers/OpenAiProviderTest.php
+++ b/tests/unit/Providers/OpenAiProviderTest.php
@@ -579,12 +579,14 @@ class OpenAiProviderTest extends TestCase {
 	public function testTextToSpeechProvider(): void {
 		$TTSProvider = new TextToSpeechProvider(
 			$this->openAiApiService,
-			$this->createMock(\OCP\IL10N::class),
+			$l10n = $this->createMock(\OCP\IL10N::class),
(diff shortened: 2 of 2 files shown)


lib/Service/WatermarkingService.php
```diff
@@ -0,0 +1,135 @@
...
+use lsolesen\pel\PelExif;
+use lsolesen\pel\PelIfd;
+use lsolesen\pel\PelJpeg;
+use lsolesen\pel\PelTag;
+use lsolesen\pel\PelTiff;
+use OCP\ITempManager;
+use Psr\Log\LoggerInterface;
+
+class WatermarkingService {
+	public const COMMENT = 'Generated with Artificial Intelligence';
```
julien-nc: Should this be translated? If so, let's use the language of the task's owner.

lib/Service/WatermarkingService.php
```diff
@@ -0,0 +1,135 @@
...
+			$text = self::COMMENT;
+
+			$img = imagecreatefromstring($image);
+			$font = 5;// built-in font 1-5
+			$white = imagecolorallocate($img, 255, 255, 255);
+			$black = imagecolorallocate($img, 0, 0, 0);
+
+			$w = imagefontwidth($font) * strlen($text);
+			$h = imagefontheight($font);
+			$px = imagesx($img) - $w - 10;
```
lukasdotcom: A smaller font size should be used if the image is too small to show all of the text. Here is an example for a 256x256 image where the text is cut off:
image

lib/TaskProcessing/TextToSpeechProvider.php
```diff
@@ -113,14 +115,18 @@ public function getOptionalOutputShapeEnumValues(): array {
...
+	public function process(?string $userId, array $input, callable $reportProgress, bool $includeWatermark = true): array {
 
 		if (!isset($input['input']) || !is_string($input['input'])) {
 			throw new RuntimeException('Invalid prompt');
 		}
 		// For OpenAI the text input limit is 4096 characters (https://platform.openai.com/docs/api-reference/audio/createSpeech#audio-createspeech-input)
 		$prompt = $input['input'];
 
+		if ($includeWatermark) {
+			$prompt .= "\n\n" . $this->l->t('This was generated using Artificial Intelligence.');
```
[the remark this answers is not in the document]
lukasdotcom: Yeah that seems very hard to do in php so we can just keep it the way you had it.

PR #311, opened 2025-12-15
feat: support multiple providers

lukasdotcom:
**Simple Overview**
Adds hidden menus for override in the ui. When you select a different context you can specify settings for that context otherwise the default will be used.
image

**Description**
- Creates preferences that allow for an override for all the connection settings for image, tts, and stt. These are used whenever the url is set for one of them. You'll see all the setters and getters for that in the OpenAiSettingsService. Also made sure to use the correct information in the OpenAiAPIService and do caching for each separate context for getModels.
- Allow in the admin settings to set an override for image, tts, stt. The settings are hidden in 3 different submenus that use the new ServiceOverridePanel Component. Then changed the getModels endpoint to store the model list in 3 separate lists models, imageModels, sttModels, and ttsModels. Also added a lot of computed values for each of the settings that can be different depending on the context.

### DIFF
--- a/lib/Service/OpenAiSettingsService.php
+++ b/lib/Service/OpenAiSettingsService.php
@@ -1203,7 +1203,7 @@ class OpenAiSettingsService {
 			$this->setRequestTimeout($adminConfig['request_timeout']);
 		}
 		if (isset($adminConfig['url'])) {
-			$this->setServiceUrl(rtrim($adminConfig['url'], " /"));
+			$this->setServiceUrl(rtrim($adminConfig['url'], ' /'));
 		}
 		if (isset($adminConfig['service_name'])) {
 			$this->setServiceName($adminConfig['service_name']);
@@ -1288,7 +1288,7 @@ class OpenAiSettingsService {
 		}
 
 		if (isset($adminConfig['image_url'])) {
-			$this->setImageServiceUrl(rtrim($adminConfig['image_url'], " /"));
+			$this->setImageServiceUrl(rtrim($adminConfig['image_url'], ' /'));
 		}
 		if (isset($adminConfig['image_service_name'])) {
 			$this->setImageServiceName($adminConfig['image_service_name']);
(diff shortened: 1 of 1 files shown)


julien-nc (commented):
* Could you describe more what the change actually is? It would make it easier to navigate in the diff.
* After switching to this branch and reloading the settings, i get 
image
Is it expected?

* It is not clear what is specific to the selected context and what is global. It might help to group what is context-specific or add an indicator next to each of those fields.
* Maybe it's more explicit or easy to follow for the users if there are 4 "service" blocks: a global one, one for image gen, one for tts, one for stt. Each block would be placed in the related section.

lib/Service/OpenAiAPIService.php
```diff
@@ -66,21 +66,47 @@ public function createQuotaUsage(string $userId, int $type, int $usage) {
...
+				return $this->l10n->t('OpenAI\'s DALL-E 2');
+			}
+			if ($serviceType === 'tts') {
+				$this->l10n->t('OpenAI\'s Text to Speech');
+			}
 			return 'OpenAI';
 		} else {
 			$serviceName = $this->openAiSettingsService->getServiceName();
+			if ($serviceType === 'image' && $this->openAiSettingsService->imageOverrideEnabled()) {
+				$serviceName = $this->openAiSettingsService->getImageServiceName();
```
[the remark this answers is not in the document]
lukasdotcom: I'm not sure. It does fallback to LocalAI if a service name isn't defined and I think that is a better default than the service name of the text processing provider.

lib/Service/OpenAiAPIService.php
```diff
@@ -1095,15 +1159,15 @@ public function request(?string $userId, string $endPoint, array $params = [], s
 	 * @return bool whether the T2I provider is available
 	 */
 	public function isT2IAvailable(): bool {
-		if ($this->isUsingOpenAi()) {
+		if ($this->isUsingOpenAi() || $this->openAiSettingsService->imageOverrideEnabled()) {
```
kyteinsky: maybe it's more likely that it would be available when it's been overridden but we can't be sure
same for others

src/components/AdminSettings.vue
```diff
@@ -760,69 +795,83 @@ export default {
 				console.error(error)
 			})
 		},
+		async getAllModels(shouldSave = true) {
+			const models = this.getModels() // for text
+			console.error(this.models)
```
kyteinsky:
```suggestion
			console.debug(this.models)
```

lib/Service/OpenAiAPIService.php
kyteinsky: cool. I meant that if a service is setup in image for example that does not support images, it should not automatically be assumed it supports it, but I suppose this is fine too, why would anyone override it otherwise.

PR #335, opened 2026-02-04
Add audio translation task type and provider

julien-nc:
* New audio2audio:translate task type (should be created in server?)
* Audio translation provider
* Factorize translation logic in a service
* Use correct user language for IL10N text translations happening in the task

### DIFF
--- a/lib/TaskProcessing/AudioToAudioTranslateProvider.php
+++ b/lib/TaskProcessing/AudioToAudioTranslateProvider.php
@@ -77,7 +77,6 @@ class AudioToAudioTranslateProvider implements IProvider, ISynchronousOptionsAwa
 		];
 	}
 
-
 	public function getOptionalInputShape(): array {
 		return [];
 	}
--- a/lib/TaskProcessing/AudioToAudioTranslateProvider.php
+++ b/lib/TaskProcessing/AudioToAudioTranslateProvider.php
@@ -41,6 +41,7 @@ class AudioToAudioTranslateProvider implements IProvider, ISynchronousOptionsAwa
 		private IL10N $l,
 		private IAppConfig $appConfig,
 		private IUserManager $userManager,
+		private ?string $userId,
 	) {
 	}
 
@@ -80,15 +81,45 @@ class AudioToAudioTranslateProvider implements IProvider, ISynchronousOptionsAwa
 	}
 
 	public function getOptionalInputShape(): array {
-		return [];
+		return [
+			'tts_voice' => new ShapeDescriptor(
+				$this->l->t('Voice'),
+				$this->l->t('The voice to use'),
+				EShapeType::Enum
+			),
+			'tts_model' => new ShapeDescriptor(
+				$this->l->t('Model'),
+				$this->l->t('The model used to generate the speech'),
+				EShapeType::Enum
+			),
+			'tts_speed' => new ShapeDescriptor(
+				$this->l->t('Speed'),
+				$this->openAiAPIService->isUsingOpenAi(Application::SERVICE_TYPE_TTS)
+					? $this->l->t('Speech speed modifier (Valid values: 0.25-4)')
+					: $this->l->t('Speech speed modifier'),
+				EShapeType::Number
+			),
+		];
 	}
 
 	public function getOptionalInputShapeEnumValues(): array {
-		return [];
+		$voices = json_decode($this->appConfig->getValueString(Application::APP_ID, 'tts_voices', lazy: true)) ?: Application::DEFAULT_SPEECH_VOICES;
+		return [
+			'tts_voice' => array_map(function ($v) {
+				return new ShapeEnumValue($v, $v);
+			}, $voices),
+			'tts_model' => $this->openAiAPIService->getModelEnumValues($this->userId, Application::SERVICE_TYPE_TTS),
+		];
 	}
 
 	public function getOptionalInputShapeDefaults(): array {
-		return [];
+		$adminVoice = $this->appConfig->getValueString(Application::APP_ID, 'default_speech_voice', lazy: true) ?: Application::DEFAULT_SPEECH_VOICE;
+		$adminModel = $this->appConfig->getValueString(Application::APP_ID, 'default_speech_model_id', lazy: true) ?: Application::DEFAULT_SPEECH_MODEL_ID;
+		return [
+			'tts_voice' => $adminVoice,
+			'tts_model' => $adminModel,
+			'tts_speed' => 1,
+		];
 	}
 
 	public function getOutputShapeEnumValues(): array {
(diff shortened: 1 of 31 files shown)


lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,195 @@
...
+		if ($includeWatermark) {
+			if ($userId !== null) {
+				$user = $this->userManager->getExistingUser($userId);
+				$lang = $this->l10nFactory->getUserLanguage($user);
+				$l = $this->l10nFactory->get(Application::APP_ID, $lang);
+				$ttsPrompt .= "\n\n" . $l->t('This was generated using Artificial Intelligence.');
+			} else {
+				$ttsPrompt .= "\n\n" . $this->l->t('This was generated using Artificial Intelligence.');
+			}
+		}
```
kyteinsky: this can also work but it would add the text/audio in the user's language which may or may not be the target language.

julien-nc: The watermark is now appended to the input transcript in the language of the user. So it ends up being in the audio and text outputs in the target language.

julien-nc:
I had to update the composer dependencies to fix CI issues.

lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,278 @@
...
+		return AudioToAudioTranslateTaskType::ID;
+	}
+
+	public function getExpectedRuntime(): int {
+		return 60;
+	}
+
+	public function getInputShapeEnumValues(): array {
+		$coreL = $this->l10nFactory->getLanguages();
+		$languages = array_merge($coreL['commonLanguages'], $coreL['otherLanguages']);
```
lukasdotcom: Use Translate Service instead otherwise #357 will be an issue

lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,278 @@
...
+				$e->getCode(),
+				$e,
+			);
+		}
+
+		$reportProgress(0.6);
+
+		// TTS
+		$ttsPrompt = $translatedText;
+		if (isset($input['model']) && is_string($input['model'])) {
```
marcelklehr: copy pasta

lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,278 @@
...
+				EShapeType::Text,
+			),
+		];
+	}
+
+	public function getOptionalOutputShapeEnumValues(): array {
+		return [];
+	}
+
+	public function process(
```
lukasdotcom, line 142: Is it possible in openai to have it translate it directly in one api call like in AudioToAudio Chat?

lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	public function getOptionalOutputShape(): array {
+		return [
+			'text_input' => new ShapeDescriptor(
+				$this->l->t('Audio transcription'),
+				$this->l->t('The transcribed audio input'),
+				EShapeType::Text,
+			),
+		];
```
marcelklehr, line 135: translated_text is missing

lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,278 @@
...
+		try {
+			$translatedText = $this->translateService->translate(
+				$transcription, $input['origin_language'], $input['target_language'],
+				$completionModel, $maxTokens, $userId,
+			);
+
+			if ($preferStreaming) {
+				$reportOutput([
+					'text_input' => $transcription,
+					'translated_text' => $translatedText,
```
marcelklehr: or this should be text_output

lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,195 @@
...
+		$voice = $this->appConfig->getValueString(Application::APP_ID, 'default_speech_voice', Application::DEFAULT_SPEECH_VOICE, lazy: true) ?: Application::DEFAULT_SPEECH_VOICE;
+		$speed = 1;
+		try {
+			$apiResponse = $this->openAiAPIService->requestSpeechCreation($userId, $ttsPrompt, $ttsModel, $voice, $speed);
+
+			if (!isset($apiResponse['body'])) {
+				$this->logger->warning('Text to speech generation failed: no speech returned');
+				throw new ProcessingException('Text to speech generation failed: no speech returned');
+			}
+			$translatedAudio = $includeWatermark ? $this->watermarkingService->markAudio($apiResponse['body']) : $apiResponse['body'];
```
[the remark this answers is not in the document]
lukasdotcom: The audio should always be marked directly as there is specific metadata that is added to the audio.

lukasdotcom (commented):
Didn't look that deeply yet, but found a few things. Also it might be worth it to use the new methods in TranslateService for the old task processing provider too.

lib/TaskProcessing/AudioToAudioTranslateProvider.php
julien-nc: Nope, wrong key, should use `text_output` which is in the task type.

lib/TaskProcessing/AudioToAudioTranslateProvider.php
julien-nc: Fixed, thanks

lib/TaskProcessing/AudioToAudioTranslateProvider.php
julien-nc: Yep, that was the mistake.

lib/TaskProcessing/AudioToAudioTranslateProvider.php
julien-nc: Sure but then it's a special case when OpenAI is used, it would only work with multimodal models (which we don't have a list of) and it relies on a system prompt that can be misinterpreted by the multimodal model.
I think it's fine like that for now. Wdyt?

marcelklehr:
A test for the AudioToAudioTranslate provider would be nice, but I'm fine if we go without :)

lib/TaskProcessing/AudioToAudioTranslateTaskType.php
```diff
@@ -0,0 +1,87 @@
...
+namespace OCA\OpenAi\TaskProcessing;
+
+use OCA\OpenAi\AppInfo\Application;
+use OCP\IL10N;
+use OCP\TaskProcessing\EShapeType;
+use OCP\TaskProcessing\ITaskType;
+use OCP\TaskProcessing\ShapeDescriptor;
+
+class AudioToAudioTranslateTaskType implements ITaskType {
+	public const ID = Application::APP_ID . ':audio2audio:translate';
```
marcelklehr, line 19: Any reason to have this not in server?

marcelklehr: (Except that server PRs are annoying :D )

lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,277 @@
...
+				$e->getCode(),
+				$e,
+			);
+		}
+		if ($includeWatermark) {
+			if ($userId !== null) {
+				$user = $this->userManager->getExistingUser($userId);
+				$lang = $this->l10nFactory->getUserLanguage($user);
+				$l = $this->l10nFactory->get(Application::APP_ID, $lang);
+				$transcription .= "\n\n" . $l->t('This was generated using Artificial Intelligence.');
```
marcelklehr: This will be translated to the target language as well. is that what we want? Or do we want the watermark to always be in english / the user language?

lib/TaskProcessing/AudioToAudioTranslateTaskType.php
julien-nc: I'll add it later and adjust here.

lib/TaskProcessing/AudioToAudioTranslateProvider.php
julien-nc: Up to debate. I think it should be the language of the user who scheduled the task.

lib/TaskProcessing/AudioToAudioTranslateProvider.php
lukasdotcom: That's fine then.

lib/TaskProcessing/AudioToAudioTranslateProvider.php
julien-nc: But @kyteinsky suggested that it would be translated to the target language.

marcelklehr (approved):

lib/TaskProcessing/AudioToAudioTranslateProvider.php
marcelklehr: Don't mind either way, I just think it should be intentional, perhaps make a comment, we can always change it later .

PR #352, opened 2026-03-04
Fix: Follow override image gen service URL

julien-nc:
* to determine if we are using OpenAI (fixes the default toggle value)
* to get the credentials when retrieving the images

Remaining issue for later: The user-defined credentials will be ignored if an admin overrides the image gen service URL.


lib/Service/OpenAiAPIService.php
```diff
@@ -884,24 +884,38 @@ public function requestImageCreation(
 	 * @return array
 	 */
 	public function getImageRequestOptions(?string $userId): array {
-		$timeout = $this->openAiSettingsService->getRequestTimeout();
 		$requestOptions = [
-			'timeout' => $timeout,
```
lukasdotcom: Timeout should always have some value. When ImageRetrieval isn't authenticated timeout will not be defined.

lukasdotcom (changes requested):
Looks good other than one small thing. Nice catch on getIsImageRetrievalAuthenticated.

PR #359, opened 2026-04-14
feat(OpenAiAPIService): Add retry logic for HTTP 429 responses

### LINKED ISSUE #354: Use exponential backoff when hitting `too many requests` responses

latest

latest

-

### Describe the Bug

When the OpenAI API responds with `Too many requests` due to an imposed rate limit that is hit, the app currently just fails the task.

### Expected Behavior

integration_openai should retry the request employing exponential backoff.

### To Reproduce

Use IONOS API and send mroe than 5 tasks per second.

marcelklehr:
fixes #354

### DIFF
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -1166,7 +1166,7 @@ class OpenAiAPIService {
 						} else {
 							$sleep = (int)$retryAfter;
 						}
-						$sleep += random_int(10, 50); // add some jitter to avoid thundering herd problem
+						$sleep += random_int(5, 30); // add some jitter to avoid thundering herd problem
 					}
 					$this->logger->warning("Rate limit exceeded, retrying in $sleep seconds", ['retry_count' => $retryCount]);
 					sleep($sleep);
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -50,6 +50,7 @@ class OpenAiAPIService {
 		private INotificationManager $notificationManager,
 		private QuotaRuleService $quotaRuleService,
 		IClientService $clientService,
+		private bool $isCLI,
 	) {
 		$this->client = $clientService->newClient();
 	}
@@ -1148,7 +1149,7 @@ class OpenAiAPIService {
 			return ['body' => $body];
 		} catch (ClientException|ServerException $e) {
 			if ($e->getResponse()->getStatusCode() === Http::STATUS_TOO_MANY_REQUESTS) {
-				if ($retryCount < 3) {
+				if ($retryCount < 3 && $this->isCLI) {
 					if (empty($e->getResponse()->getHeader('Retry-After'))) {
 						$sleep = random_int(10, 120);
 					} else {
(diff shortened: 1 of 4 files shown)


lib/Service/OpenAiAPIService.php
```diff
@@ -1145,6 +1147,16 @@ public function request(
 			}
 			return ['body' => $body];
 		} catch (ClientException|ServerException $e) {
+			if ($e->getResponse()->getStatusCode() === 429) {
+				if ($retryCount < 3) {
+					$sleep = random_int(1, 120);
```
marcelklehr:
```suggestion
					$sleep = random_int(30, 120);
```

marcelklehr:
```suggestion
					$sleep = random_int(10, 120);
```

julien-nc (approved):

lib/Service/OpenAiAPIService.php
kyteinsky: would be nice to use `retry-after` header from the response if it's available and not more than php's execution time.
IONOS also returns this, just bombarded their servers.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After

<details>

<summary>IONOS headers</summary>

```
< HTTP/2 429 
< date: Tue, 14 Apr 2026 12:50:02 GMT
< content-type: application/json; charset=utf-8
< content-length: 186
< ionos-request-id: c810ac9fc630aa35314ffcfe32a99ba2
< retry-after: 1
< strict-transport-security: max-age=31536000; includeSubDomains
< vary: Origin
< x-content-type-options: nosniff
< x-frame-options: DENY
< x-ratelimit-burst: 10
< x-ratelimit-limit: 300
< x-ratelimit-remaining: 0
< x-trace-id: b687b3c732c4714528eb2e757703cd13
< access-control-allow-credentials: true
< access-control-allow-origin: *
```

</details>

lib/Service/OpenAiAPIService.php
```diff
@@ -1145,6 +1147,33 @@ public function request(
...
+							if ($retryAfterTime !== false) {
+								$sleep = max(0, $retryAfterTime - time());
+							} else {
+								// fallback to random sleep if the header is not parsable
+								$sleep = random_int(10, 120);
+							}
+						} else {
+							$sleep = (int)$retryAfter;
+						}
+						$sleep += random_int(10, 50);
```
kyteinsky: maybe between 1 and 10 is good enough but not too sure

kyteinsky (approved):

lib/Service/OpenAiAPIService.php
```diff
@@ -1145,6 +1148,33 @@ public function request(
...
+							if ($retryAfterTime !== false) {
+								$sleep = max(0, $retryAfterTime - time());
+							} else {
+								// fallback to random sleep if the header is not parsable
+								$sleep = random_int(10, 120);
+							}
+						} else {
+							$sleep = (int)$retryAfter;
+						}
+						$sleep += random_int(10, 50); // add some jitter to avoid thundering herd problem
```
marcelklehr:
```suggestion
						$sleep += random_int(5, 30); // add some jitter to avoid thundering herd problem
```

lukasdotcom (approved):

PR #362, opened 2026-05-06
feat(TaskProcessing): add TextToTextReformatParagraphs task processing handler

lukasdotcom:
Openai integration implementation for: https://github.com/nextcloud/server/pull/60171 (It needs that branch to work)

Conserves tokens and prevents content from being lost by having the llm only generate the first sentence for each paragraph and then using code to create the final output.

PR #363, opened 2026-05-06
Create enhanced AudioToTextEnhancedProvider that also reformats text from transcription and ReformatParagraphs provider

lukasdotcom:
Built off of https://github.com/nextcloud/integration_openai/pull/362 and creates an AudioToTextEnhancedProvider that can be used instead that also reformats text. (582825fc8b9b657c1cc648543a6321cf24391dbc is the important commit)

### DIFF
--- a/lib/TaskProcessing/AudioToTextEnhancedProvider.php
+++ b/lib/TaskProcessing/AudioToTextEnhancedProvider.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\OpenAi\TaskProcessing;
+
+use OCA\OpenAi\AppInfo\Application;
+use OCA\OpenAi\Service\OpenAiAPIService;
+use OCP\IAppConfig;
+use OCP\IL10N;
+use OCP\TaskProcessing\IManager;
+use OCP\TaskProcessing\Task;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+// Built on top of the AudioToTextProvider to add paragraph reformatting.
+class AudioToTextEnhancedProvider extends AudioToTextProvider {
+	private const REFORMAT_PARAGRAPHS_TASK_TYPE_ID = 'core:text2text:reformatparagraphs';
+
+	private IManager $taskProcessingManager;
+
+	public function __construct(
+		OpenAiAPIService $openAiAPIService,
+		LoggerInterface $logger,
+		IAppConfig $appConfig,
+		IL10N $l,
+		IManager $taskProcessingManager,
+	) {
+		parent::__construct($openAiAPIService, $logger, $appConfig, $l);
+		$this->taskProcessingManager = $taskProcessingManager;
+	}
+
+	public function getId(): string {
+		return parent::getId() . '-enhanced';
+	}
+
+	public function getName(): string {
+		return parent::getName() . ' ' . $this->l->t('(with paragraph reformatting)');
+	}
+
+	public function process(?string $userId, array $input, callable $reportProgress): array {
+		$transcription = parent::process($userId, $input, $reportProgress)['output'];
+
+		$reformatTask = new Task(
+			self::REFORMAT_PARAGRAPHS_TASK_TYPE_ID,
+			['input' => $transcription],
+			Application::APP_ID,
+			$userId,
+			'audio2text_enhanced',
+		);
+
+		try {
+			$finished = $this->taskProcessingManager->runTask($reformatTask);
+			$output = $finished->getOutput();
+			if (is_array($output) && isset($output['output']) && is_string($output['output']) && $output['output'] !== '') {
+				return ['output' => $output['output']];
+			}
+			$this->logger->warning('ReformatParagraphs follow-up task returned no usable output, falling back to raw transcription');
+		} catch (Throwable $e) {
+			$this->logger->warning('ReformatParagraphs follow-up task failed, falling back to raw transcription: ' . $e->getMessage(), ['exception' => $e]);
+		}
+
+		return ['output' => $transcription];
+	}
+}
--- a/lib/TaskProcessing/AudioToTextEnhancedProvider.php
+++ b/lib/TaskProcessing/AudioToTextEnhancedProvider.php
@@ -11,43 +11,81 @@ namespace OCA\OpenAi\TaskProcessing;
 
 use OCA\OpenAi\AppInfo\Application;
 use OCA\OpenAi\Service\OpenAiAPIService;
-use OCP\IAppConfig;
-use OCP\IL10N;
 use OCP\TaskProcessing\IManager;
+use OCP\TaskProcessing\ISynchronousProvider;
 use OCP\TaskProcessing\Task;
+use OCP\TaskProcessing\TaskTypes\AudioToText;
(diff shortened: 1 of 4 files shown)


lib/TaskProcessing/AudioToTextEnhancedProvider.php
```diff
@@ -0,0 +1,70 @@
...
+use OCA\OpenAi\Service\OpenAiAPIService;
+use OCP\IAppConfig;
+use OCP\IL10N;
+use OCP\TaskProcessing\IManager;
+use OCP\TaskProcessing\Task;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+// Built on top of the AudioToTextProvider to add paragraph reformatting.
+class AudioToTextEnhancedProvider extends AudioToTextProvider {
```
marcelklehr: I think we should not have providers subclass other providers. You can inject the provider via DI and use it like that.

lib/TaskProcessing/AudioToTextEnhancedProvider.php
```diff
@@ -0,0 +1,70 @@
...
+			'audio2text_enhanced',
+		);
+
+		try {
+			$finished = $this->taskProcessingManager->runTask($reformatTask);
+			$output = $finished->getOutput();
+			if (is_array($output) && isset($output['output']) && is_string($output['output']) && $output['output'] !== '') {
+				return ['output' => $output['output']];
+			}
+			$this->logger->warning('ReformatParagraphs follow-up task returned no usable output, falling back to raw transcription');
```
marcelklehr, line 102: Good idea to fallback here!

lib/TaskProcessing/AudioToTextEnhancedProvider.php
```diff
@@ -0,0 +1,70 @@
...
+use OCP\IAppConfig;
+use OCP\IL10N;
+use OCP\TaskProcessing\IManager;
+use OCP\TaskProcessing\Task;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+// Built on top of the AudioToTextProvider to add paragraph reformatting.
+class AudioToTextEnhancedProvider extends AudioToTextProvider {
+	private const REFORMAT_PARAGRAPHS_TASK_TYPE_ID = 'core:text2text:reformatparagraphs';
```
marcelklehr: Why introduce this constant?

lib/TaskProcessing/ReformatParagraphsProvider.php
```diff
@@ -0,0 +1,203 @@
...
+
+		$maxTokens = null;
+		if (isset($input['max_tokens']) && is_int($input['max_tokens'])) {
+			$maxTokens = $input['max_tokens'];
+		}
+
+		if (isset($input['model']) && is_string($input['model'])) {
+			$model = $input['model'];
+		} else {
+			$model = $this->appConfig->getValueString(Application::APP_ID, 'default_completion_model_id', Application::DEFAULT_MODEL_ID, lazy: true) ?: Application::DEFAULT_MODEL_ID;
```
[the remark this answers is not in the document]
lukasdotcom: This could be changed in the other providers too

lib/TaskProcessing/AudioToTextEnhancedProvider.php
lukasdotcom: I really just wanted to reuse the process method so that code isn't duplicated, so DI works.

lib/TaskProcessing/AudioToTextEnhancedProvider.php
lukasdotcom: It was too make it one place to change if the server pr required that, but because it is merged I can just get the id from server directly.

lib/TaskProcessing/AudioToTextEnhancedProvider.php
```diff
@@ -0,0 +1,108 @@
...
+	public function getOptionalInputShapeDefaults(): array {
+		return $this->audioToTextProvider->getOptionalInputShapeDefaults();
+	}
+
+	public function getOutputShapeEnumValues(): array {
+		return $this->audioToTextProvider->getOutputShapeEnumValues();
+	}
+
+	public function getOptionalOutputShape(): array {
+		return $this->audioToTextProvider->getOptionalOutputShape();
```
marcelklehr: Can we hardcode these? Otherwise changes in the audioToTextProvider will automatically change this provider as well, which feels like to much magic.

lib/TaskProcessing/ReformatParagraphsProvider.php
```diff
@@ -0,0 +1,208 @@
...
+		$progress = 0.0;
+
+		foreach ($chunks as $chunk) {
+			$systemPrompt = 'Analyze the provided text and split it into paragraphs based exclusively on thematic shifts. '
+				. 'Follow these strict constraints: '
+				. 'Thematic breaks only: Do not create a new paragraph for rhythm, style, or sentence flow. '
+				. 'A break is allowed only when the subject matter changes significantly. '
+				. 'Output format: For each identified paragraph, return only the first 8 to 12 words verbatim from the input. '
+				. 'Structure: Return exactly one paragraph per line. Do not include bullets, html tags, numbering, summaries, quotes, or any additional text. '
+				. 'Single topic: If the text covers only one topic, return exactly one line.';
```
marcelklehr: Tested with output from stt_whisper2 and Mistral Small 24B from IONOS and it doesn't seem to work for me :/

marcelklehr: I get the complete blob back from the task type, same as the input

lukasdotcom: Interesting what prompt did you use. I used the one below and tested it on Mistral Small 24B and it splits it into two paragraphs:
<details>
Dogs have been humanity's most devoted companions for thousands of years, evolving from wild wolves into loyal members of our families. Renowned for their incredible intelligence, adaptability, and emotional connection, they serve in countless roles ranging from herding flocks and searching for disaster victims to providing comfort to those with anxiety. Whether they are playful puppies chasing a ball, energetic adults running off-leash, or quiet guardians watching over the home, dogs bring a unique blend of energy, affection, and unconditional love into our lives. Their ability to read human body language and emotions has fostered a deep bond between species, making them not just pets, but true partners in our daily journeys. Cats are captivating creatures that have held a unique place in human history for millennia, revered both as loyal companions and mysterious symbols of grace. With their sleek, muscular bodies and agile movements, they are masters of the hunt, capable of pouncing with pinpoint accuracy even in the dimmest light. Beyond their physical prowess, cats possess an enigmatic personality that ranges from aloof independence to devoted affection, often choosing their owners with a quiet intuition. Whether lounging lazily in a sunbeam, watching the world with wide, golden eyes, or performing a silent, graceful leap, cats embody a blend of elegance and mystery that continues to enchant people everywhere.
</details>

lib/TaskProcessing/AudioToTextEnhancedProvider.php
lukasdotcom: Some of them it makes sense to change like for the output ones, but I would say the input ones should not be changed because they are passed into the audioToTextProvider directly anyway and should always be the same.

marcelklehr: Alright, that makes sense, yes.

lib/TaskProcessing/ReformatParagraphsProvider.php
marcelklehr: I tried with the transcript of the last company call from may 5th

marcelklehr (approved):

PR #364, opened 2026-05-07
Fix/max tokens hardcoded

n-iv:
We noticed some occurrences of 'max_tokens' => 1000 in the Nextcloud codebase, specifically in:
- apps/integration_openai/lib/TaskProcessing/TextToTextProvider.php
- apps/integration_openai/lib/TaskProcessing/HeadlineProvider.php
- apps/integration_openai/lib/TaskProcessing/TopicsProvider.php

This prevents requests from being executed in full, and leads to half-generated documents or truncated answers, which ruins the user experience.

We replaced those faulty lines with:
'max_tokens' => $this->openAiSettingsService->getMaxTokens()

### DIFF
--- a/lib/TaskProcessing/HeadlineProvider.php
+++ b/lib/TaskProcessing/HeadlineProvider.php
@@ -82,7 +82,7 @@ class HeadlineProvider implements ISynchronousProvider {
 			? ($this->appConfig->getValueString(Application::APP_ID, 'default_completion_model_id', Application::DEFAULT_MODEL_ID, lazy: true) ?: Application::DEFAULT_MODEL_ID)
 			: $this->appConfig->getValueString(Application::APP_ID, 'default_completion_model_id', lazy: true);
 		return [
-			'max_tokens' => 1000,
+			'max_tokens' => $this->openAiSettingsService->getMaxTokens(),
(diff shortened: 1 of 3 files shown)


lib/TaskProcessing/HeadlineProvider.php
```diff
@@ -82,7 +82,7 @@ public function getOptionalInputShapeDefaults(): array {
 			? ($this->appConfig->getValueString(Application::APP_ID, 'default_completion_model_id', Application::DEFAULT_MODEL_ID, lazy: true) ?: Application::DEFAULT_MODEL_ID)
 			: $this->appConfig->getValueString(Application::APP_ID, 'default_completion_model_id', lazy: true);
 		return [
-			'max_tokens' => 1000,
+			'max_tokens' => $this->openAiSettingsService->getMaxTokens(),
```
lukasdotcom: This one can be set to 100 like EmojiProvider as headlines shouldn't be very long

```suggestion
			'max_tokens' => 100,
```

lukasdotcom (approved):
Thanks for the catch.

lukasdotcom:
Thank you for your contribution @n-iv

PR #365, opened 2026-05-08
Better image generation

lukasdotcom:
Improves image generation
Example: dog with hat with Nextcloud logo
Before
image

After
image

### DIFF
--- a/lib/TaskProcessing/TextToImageImprovedPromptProvider.php
+++ b/lib/TaskProcessing/TextToImageImprovedPromptProvider.php
@@ -45,6 +45,7 @@ class TextToImageImprovedPromptProvider implements ISynchronousWatermarkingProvi
 	}
 
 	public function getExpectedRuntime(): int {
+		// The text to image provider may not be openai and this assumes it is
 		return $this->textToImageProvider->getExpectedRuntime() + $this->openAiAPIService->getExpTextProcessingTime();
 	}
 
--- a/lib/TaskProcessing/TextToImageImprovedPromptProvider.php
+++ b/lib/TaskProcessing/TextToImageImprovedPromptProvider.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace OCA\OpenAi\TaskProcessing;
 
 use OCA\OpenAi\AppInfo\Application;
+use OCA\OpenAi\Service\OpenAiAPIService;
 use OCP\IL10N;
 use OCP\TaskProcessing\IManager;
 use OCP\TaskProcessing\ISynchronousWatermarkingProvider;
@@ -25,6 +26,7 @@ class TextToImageImprovedPromptProvider implements ISynchronousWatermarkingProvi
 		private IManager $taskProcessingManager,
 		private LoggerInterface $logger,
 		private IL10N $l10n,
+		private OpenAiAPIService $openAiAPIService,
 	) {
 	}
 
@@ -41,7 +43,7 @@ class TextToImageImprovedPromptProvider implements ISynchronousWatermarkingProvi
 	}
 
 	public function getExpectedRuntime(): int {
-		return $this->textToImageProvider->getExpectedRuntime();
+		return $this->textToImageProvider->getExpectedRuntime() + $this->openAiAPIService->getExpTextProcessingTime();
 	}
 
 	public function getInputShapeEnumValues(): array {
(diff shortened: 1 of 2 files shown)


lib/TaskProcessing/TextToImageImprovedPromptProvider.php
```diff
@@ -0,0 +1,118 @@
...
+	public function process(?string $userId, array $input, callable $reportProgress, bool $includeWatermark = true): array {
+		if (!isset($input['input']) || !is_string($input['input']) || trim($input['input']) === '') {
+			return $this->textToImageProvider->process($userId, $input, $reportProgress, $includeWatermark);
+		}
+
+		$originalPrompt = $input['input'];
+		$instruction
+			= 'Improve the following image-generation prompt so a text-to-image model produces a high quality, visually rich, coherent image. '
+			. 'Add concrete visual details (subject, composition, lighting, style) only when they are reasonable. '
+			. 'Keep the original intent. Return ONLY the improved prompt as a single line, no preface, no quotes, no explanations.' . "\n\n"
```
kyteinsky, line 99: ` as a single line` can be skipped maybe

lib/TaskProcessing/TextToImageImprovedPromptProvider.php
```diff
@@ -0,0 +1,118 @@
...
+		return $this->textToImageProvider->getOptionalInputShapeDefaults();
+	}
+
+	public function getOutputShapeEnumValues(): array {
+		return [];
+	}
+
+	public function getOptionalOutputShape(): array {
+		return [];
+	}
```
kyteinsky, line 84: maybe the improved prompt can be returned as well for transparency
not sure about the cases when improvement fails, empty string or the original propmt is returned then?

lib/TaskProcessing/TextToImageImprovedPromptProvider.php
kyteinsky: also this part could be summarised for the instruction prompt, don't know if it's aligned to openai models specifically or works for most models
https://developers.openai.com/cookbook/examples/multimodal/image-gen-models-prompting-guide#2-prompting-fundamentals

lib/TaskProcessing/TextToImageImprovedPromptProvider.php
lukasdotcom: Makes sense. I'd return an empty string.

lib/TaskProcessing/TextToImageImprovedPromptProvider.php
```diff
@@ -0,0 +1,118 @@
...
+	public function getName(): string {
+		return $this->l10n->t('%s (LLM-improved prompt)', [$this->textToImageProvider->getName()]);
+	}
+
+	public function getTaskTypeId(): string {
+		return TextToImage::ID;
+	}
+
+	public function getExpectedRuntime(): int {
+		return $this->textToImageProvider->getExpectedRuntime() + $this->openAiAPIService->getExpTextProcessingTime();
```
marcelklehr, line 49: May not be accurate if integration-openai is not used for text2text tasks

lukasdotcom:
It might also be worth it to have the llm generate multiple prompts if multiple images are requested otherwise all the images look very similar as the prompt is now more specific than what the user says.

marcelklehr (approved):

marcelklehr:
> It might also be worth it to have the llm generate multiple prompts if multiple images are requested otherwise all the images look very similar as the prompt is now more specific than what the user says.

mmh, let's not overengineer it, I'd say. The example you gave on the top looks great, btw. Was a good low hanging fruit after all :)

lib/TaskProcessing/TextToImageImprovedPromptProvider.php
lukasdotcom: I did try adding some of that advice, but the quality got worse so I'll leave it how it is for now. We can always change the prompt later.

lib/TaskProcessing/TextToImageImprovedPromptProvider.php
lukasdotcom: I don't think there is really a better runtime estimation we can really use and its already not that precise.

marcelklehr: ok, then let's add a comment for this

PR #367, opened 2026-05-13
Use settings service for default model instead of duplicated appConfig code

lukasdotcom:
Applying https://github.com/nextcloud/integration_openai/pull/363#discussion_r3199209568 to all providers.


julien-nc (approved):
👍

PR #372, opened 2026-05-18
Fix: throw UnknownActivityException on Notifier

lukasdotcom:


kyteinsky (approved):

lib/Notification/Notifier.php
```diff
@@ -7,8 +7,8 @@
 
 namespace OCA\OpenAi\Notification;
 
-use InvalidArgumentException;
 use OCA\OpenAi\AppInfo\Application;
+use OCP\Activity\Exceptions\UnknownActivityException;
```
nickvergessen, line 11: In the notifier you need to throw `OCP\Notification\UnknownNotificationException`

the UnknownActivityException exception is only for activity parsing 🙈

lukasdotcom: New pr to fix this at #376

PR #381, opened 2026-06-17
feat: implement generate subtitles provider

edward-ly:
Requires https://github.com/nextcloud/server/pull/61127 and https://github.com/nextcloud/assistant/pull/577

Registers the app as a new provider for the `core:audio2text:subtitles` task, reusing the transcription API to produce output as `.srt` or `.vtt` files instead of just text. Lightly tested with an MP3 of [this podcast episode](https://podcast.james.network/@linuxprepper/episodes/teaser-battlestations-old-hardware-and-unfinished-projects).

## 🤖 AI (if applicable)

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -14,6 +14,7 @@ use OCA\OpenAi\TaskProcessing\AudioToAudioChatProvider;
 use OCA\OpenAi\TaskProcessing\AudioToAudioTranslateProvider;
 use OCA\OpenAi\TaskProcessing\AudioToTextEnhancedProvider;
 use OCA\OpenAi\TaskProcessing\AudioToTextProvider;
+use OCA\OpenAi\TaskProcessing\AudioToTextSubtitlesProvider;
 use OCA\OpenAi\TaskProcessing\ChangeToneProvider;
 use OCA\OpenAi\TaskProcessing\ContextWriteProvider;
 use OCA\OpenAi\TaskProcessing\EmojiProvider;
@@ -50,6 +51,10 @@ class Application extends App implements IBootstrap {
 	public const DEFAULT_SPEECH_VOICES = [
 		'alloy', 'ash', 'ballad', 'coral', 'echo', 'fable',
 		'onyx', 'nova', 'sage', 'shimmer', 'verse'
+	];
+	public const DEFAULT_SUBTITLE_FORMAT = 'srt';
+	public const DEFAULT_SUBTITLE_FORMATS = [
+		'srt', 'vtt'
 	];
 	public const DEFAULT_DEFAULT_IMAGE_SIZE = '1024x1024';
 	public const MAX_GENERATION_IDLE_TIME = 60 * 60 * 24 * 10;
@@ -115,6 +120,9 @@ class Application extends App implements IBootstrap {
 		}
 		if ($sttProviderEnabled) {
 			$context->registerTaskProcessingProvider(AudioToTextProvider::class);
+			if (class_exists('OCP\\TaskProcessing\\TaskTypes\\AudioToTextSubtitles')) {
+				$context->registerTaskProcessingProvider(AudioToTextSubtitlesProvider::class);
+			}
 			if (class_exists('OCP\\TaskProcessing\\TaskTypes\\TextToTextReformatParagraphs')) {
 				$context->registerTaskProcessingProvider(AudioToTextEnhancedProvider::class);
 			}
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -53,7 +53,7 @@ class Application extends App implements IBootstrap {
 		'onyx', 'nova', 'sage', 'shimmer', 'verse'
 	];
 	public const DEFAULT_SUBTITLE_FORMAT = 'srt';
-	public const DEFAULT_SUBTITLE_FORMATS = [
+	public const SUPPORTED_SUBTITLE_FORMATS = [
 		'srt', 'vtt'
 	];
 	public const DEFAULT_DEFAULT_IMAGE_SIZE = '1024x1024';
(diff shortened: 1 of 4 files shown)


lib/AppInfo/Application.php
```diff
@@ -117,6 +122,9 @@ public function register(IRegistrationContext $context): void {
 		}
 		if ($sttProviderEnabled) {
 			$context->registerTaskProcessingProvider(AudioToTextProvider::class);
+			if (class_exists('OCP\\TaskProcessing\\TaskTypes\\AudioToTextSubtitles')) {
```
lukasdotcom, line 123: I think it makes sense to use a fallback provider like for AnalyzeImage so it always is supported even on older versions.

lib/TaskProcessing/AudioToTextSubtitlesProvider.php
```diff
@@ -0,0 +1,158 @@
...
+		// Maximum file size for OpenAI is 25MB. (https://developers.openai.com/api/docs/guides/speech-to-text)
+		if ($fileSize > 25 * 1000 * 1000) {
+			throw new ProcessingException('Filesize of input too large. Max is 25MB');
+		}
+
+		$fileType = $input['input']->getMimeType();
+		if (!str_starts_with($fileType, 'audio/')) {
+			throw new ProcessingException('Invalid input file type ' . $fileType);
+		}
+		if ($this->openAiAPIService->isUsingOpenAi()) {
```
lukasdotcom, line 134: It might be worth it to have slightly looser check for services that aren't using openai. Maybe just checking that it starts with `audio/`

lukasdotcom (commented):
A few comments. Also remember to fix the psalm error

lib/AppInfo/Application.php
edward-ly: I see, I do notice now there's a `AnalyzeImagesTaskType` as a backup for the server's task type, so I could do something similar here. Is that what you mean?

lib/TaskProcessing/AudioToTextSubtitlesProvider.php
edward-ly: Is that not what is happening here? The check is stricter if using OpenAI, but if not, anything starting with `audio/` should still go through.

edward-ly:
> A few comments. Also remember to fix the psalm error

Psalm is failing because https://github.com/nextcloud/server/pull/61127 hasn't been merged yet. Once that changes, it should be green again, I think.

lib/TaskProcessing/AudioToTextSubtitlesProvider.php
lukasdotcom: Oops not sure how I missed that

lib/AppInfo/Application.php
lukasdotcom: Yes exactly and then you register the provider if the server's doesn't exist. You'll also need to change some code in the provider to ensure that it returns the correct TaskType either the server's or the custom one.

lukasdotcom:
> > A few comments. Also remember to fix the psalm error
> 
> Psalm is failing because [nextcloud/server#61127](https://github.com/nextcloud/server/pull/61127) hasn't been merged yet. Once that changes, it should be green again, I think.

If that fixes it we can wait until the server's pr is merged.

lib/AppInfo/Application.php
edward-ly: Also, now that the `min-version` has already been bumped from 33 to 35, I'm not sure if a custom task type makes sense if it won't be used at all from 35 onward. We might wanna create and maintain a `stable34` branch if we actually wanna put it to use.

lukasdotcom: I forgot that the min-version was changed to 35. Then we don't need to bother with the fallback tasktype. I wouldn't create a stable34 branch for that.

lib/TaskProcessing/AudioToTextSubtitlesProvider.php
```diff
@@ -0,0 +1,158 @@
...
+
+	public function process(?string $userId, array $input, callable $reportProgress): array {
+		if (!isset($input['input']) || !$input['input'] instanceof File || !$input['input']->isReadable()) {
+			throw new ProcessingException('Invalid input file');
+		}
+
+		$fileSize = intval($input['input']->getSize());
+		// Maximum file size for OpenAI is 25MB. (https://developers.openai.com/api/docs/guides/speech-to-text)
+		if ($fileSize > 25 * 1000 * 1000) {
+			throw new ProcessingException('Filesize of input too large. Max is 25MB');
```
kyteinsky: would be nice to have this as a user-facing exception, same for the mimetype check below
see https://github.com/nextcloud/integration_openai/pull/383

lib/AppInfo/Application.php
```diff
@@ -52,6 +53,10 @@ class Application extends App implements IBootstrap {
 		'alloy', 'ash', 'ballad', 'coral', 'echo', 'fable',
 		'onyx', 'nova', 'sage', 'shimmer', 'verse'
 	];
+	public const DEFAULT_SUBTITLE_FORMAT = 'srt';
```
kyteinsky, line 55: seems unused

lib/AppInfo/Application.php
```diff
@@ -52,6 +53,10 @@ class Application extends App implements IBootstrap {
...
+	public const DEFAULT_SUBTITLE_FORMATS = [
+		'srt', 'vtt'
+	];
```
kyteinsky: nit: how does `SUPPORTED_SUBTITLE_FORMATS` sound? Default would imply it can be changed.

kyteinsky (approved):
looks good other apart from the remaining nits (didn't test though)

julien-nc:
* rebased on main and solved the conflicts
* use the unused constant
* rename DEFAULT_SUBTITLE_FORMATS to SUPPORTED_SUBTITLE_FORMATS

PR #395, opened 2026-07-02
Clamp audio duration for quota

### LINKED ISSUE #394: SQLSTATE[22003] / units column overflow during Whisper transcription

4.5.1

v33.0.6

Firefox Ubuntu Snap 152.0.1

### Describe the Bug

During audio transcription using a local Whisper backend (such as LocalAI), the background job (`cron.php`) crashes and discards the transcription result. 

In `nextcloud.log`, the following two error logs are visible:
1. `Could not create quota usage for user: test and quota type: 2. Error: An exception occurred while executing a query: SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'units' at row 1`
2. `Can not summarize call recording as no TextToTextSummary task provider is available` (This secondary warning is raised because the transcription fails/crashes before completion, and/or `llm_provider_enabled` is set to `0`).

### Root Cause Analysis

1. **Numeric Overflow via Whisper Hallucination / Silence**
In `lib/Service/OpenAiAPIService.php`, the audio duration is calculated by fetching the `end` timestamp of the last segment in the response:
```php
if (isset($response['segments'])) {
    $audioDuration = intval(round(floatval(array_pop($response['segments'])['end'])));
}
```
When local backends encounter silence or audio artifacts, they can loop and produce extremely high timestamps or return `inf` / `nan`.
- `intval(INF)` on a 64-bit PHP environment evaluates to `PHP_INT_MAX` (`9223372036854775807`).
- The database column `units` in the `oc_openai_quota_usage` table is defined as `INT` (signed 32-bit integer, max value `2147483647`) via `Version010100Date20231008173034.php`.
- Inserting `PHP_INT_MAX` into this column triggers a `SQLSTATE[22003]` numeric value out of range database error.

2. **Broken Exception Handling**
The insertion is wrapped in a try-catch block:
```php
try {
    $this->createQuotaUsage($userId ?? '', Application::QUOTA_TYPE_TRANSCRIPTION, $audioDuration);
} catch (DBException $e) {
    ...
}
```
This block fails to catch the database exception because:
- **Casing mismatch:** `DBException` is imported as `use OCP\Db\Exception as DBException;` with a lowercase `b`, while the Nextcloud Core class namespace is `OCP\DB\Exception`.
- **Class mismatch:** When the QueryBuilder fails during `executeStatement()`, it throws a Doctrine-level exception (`\Doctrine\DBAL\Exception\DriverException`), which does not inherit from `\OCP\DB\Exception`.
- The exception goes uncaught, causing the background job to crash, and the finished transcript is lost.

*Note: This unsafe `catch (DBException $e)` pattern is present in 6 locations inside `OpenAiAPIService.php` (lines 506, 575, 608, 941, 983, 1065).*

### Expected Behavior

1. **Sanitize `$audioDuration`:** The calculated audio duration should be sanitized to ensure it is finite and fits within a signed 32-bit integer.
2. **Robust Exception Catching:** Database logging failures must never crash the main application/transcription flow. The service should catch generic `\Throwable` instead of `DBException`.

### To Reproduce

1. Configure Nextcloud with the `integration_openai` app using a local Whisper backend (like LocalAI).
2. Trigger the transcription of an audio file containing background noise/silence which causes the Whisper backend to hallucinate high segment end timestamps.
3. Run `php cron.php` or wait for the background job to process the transcription.
4. The cron job will crash with the `SQLSTATE[22003]` error, and the transcript will be lost.

### Suggested Fixes

1. **Limit `$audioDuration` bounds in `OpenAiAPIService.php`:**
```php
if (!is_finite($audioDuration) || $audioDuration < 0 || $audioDuration > 2147483647) {
    $audioDuration = 0; 
}
```

2. **Catch `\Throwable` in all 6 quota logging locations inside `OpenAiAPIService.php`:**
```php
try {
    $this->createQuotaUsage(...);
} catch (\Throwable $e) {
    $this->logger->warning('Could not create quota usage: ' . $e->getMessage());
}
```

lukasdotcom:
Resolves: #394 

Not sure how often extreme numbers for the length show up, but a user reported receiving them.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -941,6 +941,7 @@ class OpenAiAPIService {
 				$audioDuration = 0;
 			}
 
+			// Audio durations higher than this can cause errors in the database: https://github.com/nextcloud/integration_openai/issues/394
 			if ($audioDuration > 2147483647) {
 				$this->logger->warning('Audio duration is greater than 2147483647 seconds: ' . $audioDuration);
 				$audioDuration = 2147483647;
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -936,6 +936,16 @@ class OpenAiAPIService {
 		if (isset($response['segments'])) {
 			$audioDuration = intval(round(floatval(array_pop($response['segments'])['end'])));
 
+			if ($audioDuration < 0) {
+				$this->logger->warning('Audio duration is less than 0: ' . $audioDuration);
+				$audioDuration = 0;
+			}
+
+			if ($audioDuration > 2147483647) {
+				$this->logger->warning('Audio duration is greater than 2147483647 seconds: ' . $audioDuration);
+				$audioDuration = 2147483647;
+			}
+
 			try {
 				$this->createQuotaUsage($userId ?? '', Application::QUOTA_TYPE_TRANSCRIPTION, $audioDuration);
 			} catch (DBException $e) {


lib/Service/OpenAiAPIService.php
```diff
@@ -936,6 +936,16 @@ public function transcribe(
...
 
+			if ($audioDuration < 0) {
+				$this->logger->warning('Audio duration is less than 0: ' . $audioDuration);
+				$audioDuration = 0;
+			}
+
+			if ($audioDuration > 2147483647) {
+				$this->logger->warning('Audio duration is greater than 2147483647 seconds: ' . $audioDuration);
+				$audioDuration = 2147483647;
+			}
```
kyteinsky, line 948: a comment would be nice here why that value instead of PHP_INT_MAX, maybe also a link to the issue?

kyteinsky (approved):

PR #398, opened 2026-07-06
docs: add Eden AI to the list of OpenAI-compatible services

MVS-source:
## Summary

Adds **Eden AI** to the README list of services that expose an OpenAI-compatible API (alongside IONOS AI Model Hub, Plusserver and MistralAI).

[Eden AI](https://www.edenai.co/) is an EU-based, GDPR-compliant unified API. Its OpenAI-compatible endpoint works with `integration_openai` out of the box:

- Service URL: `https://api.edenai.run/v3`
- `GET /v3/models` returns an OpenAI-style list (verified: `{"object":"list","data":[...]}`, 760 models, including `@eu` regional variants)
- `POST /v3/chat/completions` works with the standard OpenAI schema

This is a docs-only change (one line), no code changes. Commit is DCO signed-off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ They might work when connecting to other services, without any guarantee.
 
 Instead of connecting to the OpenAI API for these, you can also connect to a self-hosted [LocalAI](https://localai.io) instance or [Ollama](https://ollama.com/) instance
 or to any service that implements an API similar to the OpenAI one, for example:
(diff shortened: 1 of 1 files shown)


lukasdotcom:
Hi @MVS-source,

Thank you for creating this. Sadly, I've decided to close this pr for now as we can not put every single openai compatible provider on this list as it would be way too big if we did that.

PR #400, opened 2026-07-08
Add MultimodalChatWithToolsProvider

lukasdotcom:
Adds multimodal chat and also switches to enable also settings to enable different kinds of attachments. The new multimodal settings are also used for audio2audio chat and analyze image.
image
Look at https://github.com/nextcloud/assistant/pull/602 to see all relevant pr's for this

## 🤖 AI (if applicable)


lib/Service/OpenAiAPIService.php
```diff
@@ -737,32 +729,41 @@ private function buildChatCompletionRequestParams(
 						return $formattedToolCall;
 					}, $message['tool_calls']);
 				}
+				// Handle file attachments in the history
+				if (isset($message['content']) && is_array($message['content'])) {
+					$content = [];
+					foreach ($message['content'] as $item) {
+						if ($item['type'] === 'file') {
```
marcelklehr, line 766: Can we validate the array structure here before accessing the elements?

lib/Service/OpenAiAPIService.php
```diff
@@ -737,32 +729,41 @@ private function buildChatCompletionRequestParams(
...
+							$content = array_merge($content, $this->openAiFileService->buildFileContentFromId($item['file_id'], $userId, $item['ocp_task_id'] ?? null));
```
marcelklehr: $userId may be null here

lib/Service/OpenAiFileService.php
```diff
@@ -0,0 +1,347 @@
...
+	 * @return list<array<string, mixed>> Content parts suitable for OpenAI chat message content.
+	 * @throws ProcessingException
+	 * @throws UserFacingProcessingException
+	 */
+	public function buildFileContentFromFile(?File $file): array {
+		if (!$file instanceof File || !$file->isReadable()) {
+			throw new ProcessingException('File is not readable');
+		}
+		// Maximum file size for openai is 50MB.
+		if ($this->isUsingOpenAi() && $file->getSize() > self::MAX_FILE_SIZE_BYTES) {
```
marcelklehr, line 122: Should we add a file size limit fro the general case as well?

lib/Migration/Version040600Date20260708151000.php
```diff
@@ -0,0 +1,44 @@
...
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure
+	 * @param array $options
+	 */
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+		$unset = '__unset__';
+		$newValue = $this->appConfig->getValueString(Application::APP_ID, 'multimodal_image_enabled', $unset);
+		if ($newValue !== $unset) {
+			return;
```
marcelklehr, line 34: This shifts the default state from 0 to 1, I think. Ideally, if it has been unset previously, we should set it to what used to be the default.

lib/TaskProcessing/MultimodalChatWithToolsProvider.php
```diff
@@ -0,0 +1,244 @@
...
+		if (count($returnValue['messages']) > 0 || count($returnValue['tool_calls']) > 0 || count($returnValue['images']) > 0 || count($returnValue['audio_messages']) > 0) {
+			$endTime = time();
+			$this->openAiAPIService->updateExpTextProcessingTime($endTime - $startTime);
+			$attachments = [];
+
+			// Handle image output
+			foreach ($returnValue['images'] as $image) {
+				if ($image['type'] === 'image_url') {
+					$url = $image['image_url']['url'];
+					$base64Str = explode(',', $url)[1] ?? '';
```
marcelklehr: Should we rather throw if the url is malformed?

lib/Service/OpenAiFileService.php
lukasdotcom: I can do the exact same limit as for openai unless you think a different one is better.

lib/Migration/Version040600Date20260708151000.php
lukasdotcom: Back when analyze image was created I don't think IONOS had a model with support for image input, but now that they do I think it might be worth it to change the default to on instead of off.

lib/Service/OpenAiAPIService.php
lukasdotcom: Not sure how to handle it for the fileId only path, but if it is from a task the getUserTask accepts a null value for userId.

lib/Service/OpenAiAPIService.php
lukasdotcom: Added some validation.

lib/TaskProcessing/MultimodalChatWithToolsProvider.php
lukasdotcom: Added a throw for when explode doesn't work and for when an invalid image_url is found.

julien-nc (commented):
My tests worked great with images using `google/gemini-3.1-flash-lite-image` (Nano Banana 2 Lite) on OpenRouter.

I tried using `gpt-audio` on OpenRouter as well and it could take audio as input but never generated any audio output. The model claimed it had generated some audio but in the task output: `"output_attachments":[]`.

Could this be an issue in this implementation or you think the model is just not outputting any audio content?

lukasdotcom:
> My tests worked great with images using `google/gemini-3.1-flash-lite-image` (Nano Banana 2 Lite) on OpenRouter.
> 
> I tried using `gpt-image` on OpenRouter as well and it could take images as input but never generated any audio output. The model claimed it had generated some audio but in the task output: `"output_attachments":[]`.
> 
> Could this be an issue in this implementation or you think the model is just not outputting any audio content?

Are you sure `gpt-image` can generate audio. It sounds more like a model for image generation :). I'll test it's image generation though. It might pass the images it returns in a different format than gemini. Also I'm not sure that I tested audio as I copied the code from the audio chat's implementation.

julien-nc:
@lukasdotcom I meant gpt-audio and audio input/output, sorry, typo. I edited my comment.

lib/TaskProcessing/AnalyzeImagesProvider.php
```diff
@@ -120,68 +119,7 @@ public function process(
...
-			);
-		}
-		$fileSize = 0;
-		foreach ($input['images'] as $image) {
-			if (!$image instanceof File || !$image->isReadable()) {
-				throw new ProcessingException('Invalid input file');
-			}
-			$fileSize += intval($image->getSize());
-			// Maximum file size for openai is 50MB. Seems reasonable enough to enforce for all apis though. (https://platform.openai.com/docs/guides/images-vision?api-mode=responses&format=url#image-input-requirements)
-			if ($fileSize > 50 * 1000 * 1000) {
```
julien-nc, line 139: We don't check the cumulative file size anymore. `OpenAiFileService::buildFileContentFromFile` checks if individual files are exceeding the size limit.

lib/TaskProcessing/MultimodalChatWithToolsProvider.php
```diff
@@ -0,0 +1,244 @@
...
+			throw new ProcessingException('OpenAI/LocalAI request failed: ' . $e->getMessage());
+		}
+		if (count($returnValue['messages']) > 0 || count($returnValue['tool_calls']) > 0 || count($returnValue['images']) > 0 || count($returnValue['audio_messages']) > 0) {
+			$endTime = time();
+			$this->openAiAPIService->updateExpTextProcessingTime($endTime - $startTime);
+			$attachments = [];
+
+			// Handle image output
+			foreach ($returnValue['images'] as $image) {
+				if ($image['type'] === 'image_url') {
```
julien-nc, line 216: Is this always inline data? Why is the attr called `image_url` if we get b64 data? Weird.

marcelklehr: I think it's a data URI, but could probably also be a real URL? I dunno

lukasdotcom: I only received data URI's in my testing, but I could add support for real urls.

lukasdotcom:
> @lukasdotcom I meant gpt-audio and audio input/output, sorry, typo. I edited my comment.

I think the issue is that you need to pass an extra parameter, not like with the image models that don't require anything else, but not sure if we should even duplicate the code form AudioChat for multimodal chat.
```php
$extraParams = [
			'modalities' => ['text', 'audio'],
			'audio' => ['voice' => $outputVoice, 'format' => 'mp3'],
		];
```

lib/TaskProcessing/MultimodalChatWithToolsProvider.php
julien-nc: Let's just add a comment there mentioning so far only data URIs have been seen with OpenRouter and OpenAI.

julien-nc (approved):

PR #409, opened 2026-07-22
Fix: mistral passes references when tool calls are used

### LINKED ISSUE #408: Typed Mistral message.content chunks are ignored, causing 'No result' errors

4.5.1

33.0.6.2, using the `nextcloud:33.0.6-apache` Docker image.

Not browser-specific. Observed through the Nextcloud Assistant web interface, but the failure occurs in server-side response normalization.

### Describe the Bug

When an OpenAI-compatible endpoint returns assistant `message.content` as a list of typed content chunks, integration_openai 4.5.1 ignores the content and can eventually report:

```text
No result in OpenAI/LocalAI response.
```

I encountered this repeatedly with Mistral Small 4 (`mistral-small-2603`) through `https://api.eu.mistral.ai/v1`. The failures occurred in `core:text2text:chatwithtools` after an external MCP web-search tool had completed successfully. Both single and parallel tool-use cases were observed, so parallelism was not the cause.

The relevant parser in `OpenAiAPIService::createChatCompletion()` only accepts content when it is a string:

```php
if (isset($choice['message']['content']) && is_string($choice['message']['content'])) {
    $completions['messages'][] = $choice['message']['content'];
}
```

Mistral's assistant-message schema permits content to be either a string or a list of typed chunks. Observed chunk types included:

- `text`
- `reference`
- `thinking`

Official schema references:

- [AssistantMessage](https://github.com/mistralai/client-python/blob/c94f59514e59f6341a00d92e43c6ed62566d5eeb/src/mistralai/client/models/assistantmessage.py)
- [ContentChunk](https://github.com/mistralai/client-python/blob/c94f59514e59f6341a00d92e43c6ed62566d5eeb/src/mistralai/client/models/contentchunk.py)
- [ThinkChunk](https://github.com/mistralai/client-python/blob/c94f59514e59f6341a00d92e43c6ed62566d5eeb/src/mistralai/client/models/thinkchunk.py)
- [DeltaMessage](https://github.com/mistralai/client-python/blob/c94f59514e59f6341a00d92e43c6ed62566d5eeb/src/mistralai/client/models/deltamessage.py)

#### Confirmed response metadata

I reconstructed one previously failed synthetic request and replayed it three times with the same model, messages, tools, `n: 1`, `max_tokens: 5000`, no extra parameters, and no explicit `reasoning_effort`.

All three requests returned HTTP 200, one choice, `finish_reason: stop`, no tool call, and no audio:

| Replay | `message.content` | Chunk types | Parser result |
| --- | --- | --- | --- |
| 1 | string | n/a | accepted |
| 2 | array | alternating `text` and `reference`, ending in `text` | rejected |
| 3 | array | alternating `text` and `reference`, ending in `text` | rejected |

This shows why the problem is intermittent: the same request can return either a string or typed content.

A separate public synthetic request using `reasoning_effort: "high"` returned a `thinking` chunk followed by a `text` chunk. Version 4.5.1 rejected that response as well. This was only a deterministic response-shape probe; I am not proposing `reasoning_effort` as a production setting or workaround.

No prompts, tool-result bodies, response text, credentials, user identifiers, or raw provider responses have been retained.

#### Tested local workaround for 4.5.1

I locally patched version 4.5.1 to:

- preserve string content unchanged;
- concatenate top-level `type: "text"` chunks in order;
- recognize but not expose `thinking` chunks in the 4.5.1 provider;
- ignore `reference`, unknown, null, empty, and malformed chunks safely; and
- leave outgoing requests, tool calls, and audio handling unchanged.

This fixed the observed failure. After deployment:

- an explicit web-search request completed two parallel MCP searches and returned an answer; and
- retrying an existing conversation that had previously failed with this parser error returned an answer.

The patch below is the exact tested 4.5.1 workaround. It is not necessarily the preferred implementation for current `main`, because current `main` also has a streaming parser and structured reasoning support.

<details>
<summary>Tested integration_openai 4.5.1 workaround diff</summary>

```diff
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -708,8 +708,9 @@
 			}
 
 			// always try to get a message
-			if (isset($choice['message']['content']) && is_string($choice['message']['content'])) {
-				$completions['messages'][] = $choice['message']['content'];
+			$content = $this->normalizeChatMessageContent($choice['message']['content'] ?? null);
+			if ($content !== null) {
+				$completions['messages'][] = $content;
 			}
 			if (isset($choice['message']['audio'], $choice['message']['audio']['data']) && is_string($choice['message']['audio']['data'])) {
 				$completions['audio_messages'][] = $choice['message'];
@@ -720,6 +721,36 @@
 	}
 
 	/**
+	 * Normalize OpenAI-compatible string content and typed content chunks.
+	 *
+	 * Thinking chunks are recognized but intentionally not exposed by this
+	 * provider version. Unknown and malformed chunks are ignored.
+	 */
+	private function normalizeChatMessageContent(mixed $content): ?string {
+		if (is_string($content)) {
+			return $content;
+		}
+		if (!is_array($content) || !array_is_list($content)) {
+			return null;
+		}
+
+		$textChunks = [];
+		foreach ($content as $chunk) {
+			if (!is_array($chunk)) {
+				continue;
+			}
+			if (($chunk['type'] ?? null) === 'thinking') {
+				continue;
+			}
+			if (($chunk['type'] ?? null) === 'text' && isset($chunk['text']) && is_string($chunk['text'])) {
+				$textChunks[] = $chunk['text'];
+			}
+		}
+
+		return $textChunks === [] ? null : implode('', $textChunks);
+	}
+
+	/**
 	 * @param string $configKey
 	 * @return array|null
 	 */
```

</details>

Local regression fixtures covered:

- existing OpenAI string content;
- Mistral `thinking` plus `text` chunks;
- interleaved `text` and `reference` chunks;
- multiple text and unknown chunks;
- null, empty, and malformed content; and
- unchanged existing tool-call and audio normalization.

The patched file passed PHP 8.3 lint.

Current upstream `main` also has streaming response parsing. A complete upstream fix may therefore need the same typed-content support in both non-streaming normalization and streaming SSE parsing.

#### Investigation assistance

I encountered this bug multiple times. OpenAI `gpt-5.6-sol` was used to help compare the synthetic occurrences, construct a privacy-safe replay, inspect the relevant schemas, and prepare the local workaround and tests.

I supplied the provider credential locally through `getpass`, approved and performed the production deployment, and manually validated the resulting Assistant and MCP behavior.

### Expected Behavior

OpenAI-compatible assistant responses should produce a final text result when `message.content` is either:

- a normal string; or
- a list containing valid `type: "text"` chunks.

For typed content, the text chunks should be concatenated in order. Thinking, reference, unknown, or malformed chunks should not cause otherwise valid final text to be discarded.

Existing behavior for string content, separate `reasoning_content`, tool calls, audio, and ordinary OpenAI-compatible providers should remain unchanged.

#### Intermittent provider reproduction

1. Configure integration_openai 4.5.1 with the Mistral EU OpenAI-compatible endpoint and `mistral-small-2603`.
2. Submit a sufficiently involved chat request, optionally with tools.
3. Allow the provider to complete the request successfully.
4. Repeat if necessary because Mistral may return either string or typed content for the same request.
5. When `message.content` is an array of typed chunks, observe that integration_openai returns no message and the task can fail with:

   ```text
   No result in OpenAI/LocalAI response.
   ```

#### Provider-level typed-content reproduction

A near-deterministic way to request the relevant response shape is to send a separate diagnostic request with `reasoning_effort: "high"`:

```json
{
  "model": "mistral-small-2603",
  "messages": [
    {
      "role": "user",
      "content": "Tell me more about Amsterdam:\n1. What is its history?\n2. What are its demographics?\n3. Is it worth visiting for sightseeing?"
    }
  ],
  "n": 1,
  "max_tokens": 5000,
  "reasoning_effort": "high"
}
```

In my test, the equivalent synthetic request returned HTTP 200 with:

```text
message.content type: array
chunk types: thinking, text
finish_reason: stop
```

Version 4.5.1 extracted no message from that response. The parameter is only intended to make the typed response shape reproducible; it is not proposed as a production setting or workaround.

#### Deterministic parser-level reproduction

Pass a successful synthetic chat-completion response shaped like this through the 4.5.1 response normalizer:

```json
{
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": [
          {
            "type": "thinking",
            "thinking": [
              {
                "type": "text",
                "text": "Synthetic reasoning"
              }
            ]
          },
          {
            "type": "text",
            "text": "Synthetic final answer"
          }
        ]
      },
      "finish_reason": "stop"
    }
  ]
}
```

Expected result:

```text
Synthetic final answer
```

Actual 4.5.1 result:

```text
No message is extracted.
```

lukasdotcom:
Mistral has a weird format when tool call results exist.

Resolves: #408 

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Service/OpenAiAPIService.php
+++ b/lib/Service/OpenAiAPIService.php
@@ -1474,8 +1474,19 @@ class OpenAiAPIService {
 				}
 			}
 
-			if (isset($choice['message']['content']) && is_string($choice['message']['content'])) {
-				$completions['messages'][] = $choice['message']['content'];
+			if (isset($choice['message']['content'])) {
+				if (is_string($choice['message']['content'])) {
+					$completions['messages'][] = $choice['message']['content'];
+				} elseif (is_array($choice['message']['content'])) {
+					$messageContent = [];
+					// Handles more complex mistral message content (TODO: missing reasoning for now look at https://github.com/nextcloud/integration_openai/pull/409#discussion_r3638108824)
+					foreach ($choice['message']['content'] as $content) {
+						if ($content['type'] === 'text' && is_string($content['text'])) {
+							$messageContent[] = $content['text'];
+						}
+					}
+					$completions['messages'][] = implode('', $messageContent);
+				}
 			}
 			if (isset($choice['message']['reasoning_content']) && is_string($choice['message']['reasoning_content'])) {
 				$completions['reasoning_messages'][] = $choice['message']['reasoning_content'];
--- a/tests/unit/Providers/OpenAiProviderTest.php
+++ b/tests/unit/Providers/OpenAiProviderTest.php
@@ -1010,4 +1010,90 @@ TEXT;
 		$this->quotaUsageMapper->deleteUserQuotaUsages(self::TEST_USER1);
 	}
 
+	/**
+	 * Mistral may return message.content as alternating text/reference chunks
+	 * instead of a plain string. See https://github.com/nextcloud/integration_openai/issues/408
+	 */
+	public function testCreateChatCompletionExtractsMistralTypedTextContent(): void {
+		$response = json_encode([
+			'id' => 'f984ba9c1b34430d8935ccc3263b218c',
+			'created' => 1784753166,
+			'model' => 'mistral-small-2603',
+			'usage' => [
+				'prompt_tokens' => 9,
+				'total_tokens' => 21,
+				'completion_tokens' => 12,
+				'prompt_tokens_details' => [
+					'cached_tokens' => 0,
+				],
+			],
+			'object' => 'chat.completion',
+			'choices' => [
+				[
+					'index' => 0,
+					'finish_reason' => 'stop',
+					'message' => [
+						'role' => 'assistant',
+						'tool_calls' => null,
+						'content' => [
+							[
+								'type' => 'text',
+								'text' => 'Amsterdam was founded around 1000 CE',
+							],
+							[
+								'type' => 'reference',
+								'reference_ids' => [1, 3],
+							],
+							[
+								'type' => 'text',
+								'text' => '. It has about 1.2 million residents',
+							],
+							[
+								'type' => 'reference',
+								'reference_ids' => [2],
+							],
+							[
+								'type' => 'text',
+								'text' => '. It is worth visiting.',
+							],
+						],
+					],
+				],
+			],
+		]);
+
+		$url = self::OPENAI_API_BASE . 'chat/completions';
+		$options = ['timeout' => Application::OPENAI_DEFAULT_REQUEST_TIMEOUT, 'headers' => ['User-Agent' => Application::USER_AGENT, 'Authorization' => self::AUTHORIZATION_HEADER, 'Content-Type' => 'application/json']];
+		$options['body'] = json_encode([
+			'model' => Application::DEFAULT_COMPLETION_MODEL_ID,
+			'messages' => [['role' => 'user', 'content' => 'Tell me more about Amsterdam']],
+			'n' => 1,
+			'stream' => false,
+			'max_completion_tokens' => Application::DEFAULT_MAX_NUM_OF_TOKENS,
+			'user' => self::TEST_USER1,
+		]);
+
+		$iResponse = $this->createMock(\OCP\Http\Client\IResponse::class);
+		$iResponse->method('getBody')->willReturn($response);
+		$iResponse->method('getStatusCode')->willReturn(200);
+		$iResponse->method('getHeader')->with('Content-Type')->willReturn('application/json');
+
+		$this->iClient->expects($this->once())->method('post')->with($url, $options)->willReturn($iResponse);
+
+		$result = $this->openAiApiService->createChatCompletion(
+			self::TEST_USER1,
+			Application::DEFAULT_MODEL_ID,
+			'Tell me more about Amsterdam',
+		);
+
+		$this->assertSame(
+			['Amsterdam was founded around 1000 CE. It has about 1.2 million residents. It is worth visiting.'],
+			$result['messages'],
+		);
+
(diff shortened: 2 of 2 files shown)


lib/Service/OpenAiAPIService.php
```diff
@@ -1474,8 +1474,19 @@ private function normalizeChatCompletionResponse(array $response): array {
...
+				if (is_string($choice['message']['content'])) {
+					$completions['messages'][] = $choice['message']['content'];
+				} elseif (is_array($choice['message']['content'])) {
+					$messageContent = [];
+					// Handle more complex mistral message content
+					foreach ($choice['message']['content'] as $content) {
+						if ($content['type'] === 'text' && is_string($content['text'])) {
+							$messageContent[] = $content['text'];
+						}
+					}
```
kyteinsky: looks like reasoning and multimodal output is also in this same content
maybe reasoning can be also extracted here, and then the multimodal output in the other PR?
not sure how complex it would get though

[the long and precise link to docs](https://docs.mistral.ai/api/endpoint/chat?property=operation-chat_completion_v1_chat_completions_post_responses_200_application-json_choices__operation-chat_completion_v1_chat_completions_post_responses_200_application-json_choices_chatcompletionchoice__operation-chat_completion_v1_chat_completions_post_responses_200_application-json_choices_chatcompletionchoice_message__operation-chat_completion_v1_chat_completions_post_responses_200_application-json_choices_chatcompletionchoice_message_assistantmessage__operation-chat_completion_v1_chat_completions_post_responses_200_application-json_choices_chatcompletionchoice_message_assistantmessage_content#operation-chat_completion_v1_chat_completions_post_responses_200_application-json_choices)

kyteinsky: the current changes are good to merge though, if reasoning is to be extracted later on :+1:

lukasdotcom: I added a comment that mentions a todo for the future

kyteinsky (approved):

PR #423, opened 2026-08-07
Add image editing support

lukasdotcom:
Add support for image editing. I'm waiting for the server's task type until v36.
image

Annoyingly there doesn't seem to be any standard way for image editing, so I wrote a custom handler for Openai, Openrouter, IONOS, and localAI and tested them all individually.

## 🤖 AI (if applicable)


lib/Service/OpenAiAPIService.php
```diff
@@ -936,6 +936,197 @@ public function requestImageCreation(
...
+		string $prompt,
+		array $images,
+		string $model,
+		string $size = Application::DEFAULT_DEFAULT_IMAGE_SIZE,
+	): array {
+		if ($this->isQuotaExceeded($userId, Application::QUOTA_TYPE_IMAGE, $service)) {
+			throw new Exception($this->l10n->t('Image generation quota exceeded'), Http::STATUS_TOO_MANY_REQUESTS);
+		}
+
+		$apiModel = $this->modelParam($service, $model, Application::DEFAULT_IMAGE_MODEL_ID) ?? $model;
```
marcelklehr: modelParam() returns null to mean "omit the model parameter" — every other caller honors that (requestImageCreation() ~line 944, requestTranscription() ~line 818 both guard with if ($modelParam !== null)). The ?? $model here turns the sentinel back into the literal string 'Default', and all four edit paths then set 'model' => $model unconditionally.

lib/Service/ServiceConfig.php
```diff
@@ -273,6 +273,10 @@ public function isUsingOpenRouter(): bool {
 		return str_starts_with(strtolower($this->url), 'https://openrouter.ai');
 	}
 
+	public function isUsingIonos(): bool {
+		return (bool)preg_match('#^https://([a-zA-Z0-9-]+\.)+ionos\.com#', strtolower($this->getRequestUrl()));
```
marcelklehr: - https://api.ionos.community/v1 → matches (false positive, .com inside .community)
- https://api.ionos.com.attacker.example/v1 → matches (false positive)
- https://ionos.com/v1 → does not match (the group requires a subdomain label)

Each takes the wrong branch in requestImageEdit().

lib/Service/OpenAiAPIService.php
```diff
@@ -936,6 +936,197 @@ public function requestImageCreation(
...
+
+		if ($service->isUsingOpenAi()) {
+			$apiResponse = $this->requestOpenAiImageEdit($userId, $service, $prompt, $images, $apiModel, $size);
+		} elseif ($service->isUsingOpenRouter()) {
+			$apiResponse = $this->requestOpenRouterImageEdit($userId, $service, $prompt, $images, $apiModel, $size);
+		} elseif ($service->isUsingIonos()) {
+			$apiResponse = $this->requestIonosImageEdit($userId, $service, $prompt, $images, $apiModel, $size);
+		} else {
+			$apiResponse = $this->requestLocalAiImageEdit($userId, $service, $prompt, $images, $apiModel, $size);
```
marcelklehr: requestLocalAiImageEdit() is the else for every non-OpenAI/OpenRouter/IONOS service. It POSTs images/generations with LocalAI's ref_images field. vLLM, Ollama gateways, LiteLLM etc. ignore unknown JSON fields and return a fresh image generated from the prompt alone — HTTP 200, b64_json present, provider watermarks it and reports success. User gets a plausible result that ignored their input image, nothing logged. Multipart images/edits (the requestOpenAiImageEdit shape) is the safer default; reserve ref_images for a detected LocalAI.

lib/TaskProcessing/ImageToImageProvider.php
```diff
@@ -0,0 +1,226 @@
...
+				null,
+				$this->l->t('Cannot use more than 16 input images.'),
+			);
+		}
+
+		foreach ($input['input'] as $inputFile) {
+			if (!$inputFile instanceof File || !$inputFile->isReadable()) {
+				throw new ProcessingException('Invalid input file');
+			}
+			if ($inputFile->getSize() > self::MAX_FILE_SIZE_BYTES) {
```
marcelklehr, line 156: 25 MB checked per file (line 140) with up to 16 files. Siblings cap the total: AnalyzeImagesProvider.php:129-143 and ImageToTextOcrProvider.php:119-133 both array_reduce to 50 MB. Worst case 16×25 MB → 400 MB of content, +~533 MB base64, +~533 MB from json_encode($params) — over 1.4 GB live, which fatals the worker under a default 512 MB memory_limit instead of returning a task error.

lib/TaskProcessing/ImageToImageProvider.php
```diff
@@ -0,0 +1,226 @@
...
+
+	public function getName(): string {
+		return $this->buildProviderName();
+	}
+
+	public function getTaskTypeId(): string {
+		if (class_exists('OCP\\TaskProcessing\\TaskTypes\\ImageToImage')) {
+			return \OCP\TaskProcessing\TaskTypes\ImageToImage::ID;
+		}
+		return ImageToImageTaskType::ID;
```
marcelklehr, line 61: returns openai:image2image when OCP\TaskProcessing\TaskTypes\ImageToImage is missing and core:image2image when present (since 36.0.0). Since ProviderIdentity::buildProviderId() slugifies the task type into the provider ID, the ID changes on upgrade and orphans the admin's provider preference — exactly what Version060000Date20260914120000 was written to prevent for the other task types. Also, tasks already stored with taskTypeId = 'openai:image2image' become unhandleable once the core type appears. Worth a deliberate call (possibly a follow-up migration) before this ships next to the v36 task type.

lib/Service/ServiceConfig.php
lukasdotcom: IMO it doesn't really matter right now because it just changes the format, but I made the url as strict as for the other providers.

lib/Service/OpenAiAPIService.php
lukasdotcom: Added a detection for localai that hits an endpoint that only exists for localai instances. If that endpoint gives a 200 response than locali is used.

lib/TaskProcessing/ImageToImageProvider.php
lukasdotcom: slugifyTaskType now also removes the app id if that is at the start as those task types are just overrides anyway for a core task type.

lib/TaskProcessing/ImageToImageProvider.php
```diff
@@ -0,0 +1,242 @@
...
+		$size = $this->service->getDefaultImageSize();
+		if (isset($input['size']) && is_string($input['size']) && preg_match('/^\d+x\d+$/', $input['size'])) {
+			$size = trim($input['size']);
+		}
+		if (preg_match('/^\d+x\d+$/', $size) !== 1) {
+			$size = Application::DEFAULT_DEFAULT_IMAGE_SIZE;
+		}
+		[$x, $y] = explode('x', $size, 2);
+		if ((int)$x > 4096 || (int)$y > 4096) {
+			throw new UserFacingProcessingException('size is out of bounds', userFacingMessage: $this->l->t('Cannot generate images larger than 4096x4096'));
```
marcelklehr: We could improve the validation here. Apparently, IONOS needs both dims ≤ 2048 and multiples of 16; OpenAI gpt-image-1* accepts only 1024x1024, 1024x1536, 1536x1024, auto.  Right now, size=2048x2048, or an admin default_image_size of 512x512, passes validation then 400s with no hint about size.

marcelklehr (approved):

PR #424, opened 2026-08-07
Fix mistral for pdf documents

lukasdotcom:
~Also had an llm refactor the isUsing functions into openAiSettingsService so there weren't circular dependencies.~ Thanks @cbcoutinho for finding this issue originally.

## 🤖 AI (if applicable)


edward-ly (changes requested):
Would you be able to split the refactors and the actual fix into separate commits? That would make things easier to review.

lukasdotcom:
> Would you be able to split the refactors and the actual fix into separate commits? That would make things easier to review.

Rebased it and split it into two commits.

Edit: Due to #431 the refactor isn't needed anymore.

edward-ly (approved):
Haven't tested this myself yet as I don't have a Mistral account, but the code looks fine regardless 🚀

lukasdotcom:
> Haven't tested this myself yet as I don't have a Mistral account, but the code looks fine regardless 🚀

For future reference you can actually create a free account. Mistral has a free tier for testing.

cbcoutinho:
Hi @lukasdotcom can we get this backported to v24? It looks like this app's minimum app version is 35

lukasdotcom:
> Hi @lukasdotcom can we get this backported to v24? It looks like this app's minimum app version is 35

Hi @cbcoutinho,

This can't be backported as the fix is only for multimodal chat which only exists in the main branch not any of the stable branches.

PR #432, opened 2026-09-09
fix: Cap quota period length to 10 years

### LINKED ISSUE #434: Verify inputs from admin setting

NC-Assistant: 3.5.0 / OpenAI-Local int: 4.5.2

34.0.2

Chrome

### Describe the Bug

I've entered a too large number into a field, which caused a server error, as the field could no longer be interpretet:
See the bug I've raised here: https://help.nextcloud.com/t/ai-assistant-entered-a-wrong-value-internal-server-error/249033

Request:
Check the input values for the validity, prior saving it to the database.

### Expected Behavior

not allowing me to enter something wrong

### To Reproduce

in the admin settings of the Assistent:
Enter Reset quota every: 99999999999999999
-> the admin settings are no longer accessible.

Not sure if other fiels are affected-

lukasdotcom:
Note this is done in the getter also to make sure that invalid saved values are capped.
Resolves: #434 

## 🤖 AI (if applicable)

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -62,6 +62,8 @@ class Application extends App implements IBootstrap {
 	public const MIN_CHUNK_SIZE = 500;
 	public const DEFAULT_MAX_NUM_OF_TOKENS = 1000;
 	public const DEFAULT_QUOTA_PERIOD = 30;
+	public const MAX_QUOTA_PERIOD_DAYS = 3650;
+	public const MAX_QUOTA_PERIOD_MONTHS = 120;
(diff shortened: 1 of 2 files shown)


lib/AppInfo/Application.php
```diff
@@ -62,6 +62,8 @@ class Application extends App implements IBootstrap {
 	public const MIN_CHUNK_SIZE = 500;
 	public const DEFAULT_MAX_NUM_OF_TOKENS = 1000;
 	public const DEFAULT_QUOTA_PERIOD = 30;
+	public const MAX_QUOTA_PERIOD_DAYS = 3650;
```
edward-ly, line 65: Technically, 2.5 days should be added to account for leap years, but then again, I don't know if it matters too much here.

edward-ly (approved):

PR #433, opened 2026-09-10
feat: make summary system prompt configurable

### LINKED ISSUE #405: Add admin-configurable summary system prompt (SummaryProvider currently hardcodes it)
### Describe the feature you'd like to request

`apps/integration_openai/lib/TaskProcessing/SummaryProvider.php` currently hardcodes the system prompt used for text-summarisation tasks:

```
You are a helpful assistant that summarizes text in the same language as the text.
You should only return the summary without any additional information.
```

There is no admin UI or `config:app:set` key to override this prompt. The admin-visible **Chat User Instructions for Chat Completions** field applies only to the interactive chat-completion path, not to summarisation (which goes through TaskProcessing → `SummaryProvider::process()` on a separate code path). The `max_tokens = 1000` default on this same code path is also very restrictive.

Combined, this produces very short generic summaries. Talk call-recording summaries in particular miss most of the value of the transcript — no structure, no attendees, no action items, no decisions, no next steps.

Every deployment that wants better summaries — richer meeting summaries, per-team wording, per-language variants, non-English house style, industry-specific vocabulary — has to patch PHP directly. That patch is clobbered by every app update and has to be manually re-applied. This is fragile across a fleet and is a real papercut for any admin running `integration_openai` against a LocalAI-compatible endpoint (LiteLLM, Ollama, self-hosted OpenAI-shape LLMs).

**Environment where this was observed:**
- Nextcloud 33.0.6 (Hub 26 Winter)
- integration_openai 4.5.1
- Talk 23.0.8
- Backend: LiteLLM proxy fronting local LLM (Qwen 2.5) — `url` config points at the LiteLLM endpoint

### Describe the solution you'd like

Add a **Summary system prompt** text field to `Admin → Connected accounts → OpenAI and LocalAI integration` (or wherever fits the app's existing settings UI). The field should:

- Persist to `integration_openai` app config under a stable key, e.g. `summary_system_prompt`
- Default to the current hardcoded string, so upgrade is a no-op for existing deployments
- Be read by `SummaryProvider::process()` in place of the hardcoded string; fall back to the current default when the config key is empty
- Support multi-line input (textarea, not single-line)
- Include short help text explaining the field is used verbatim as the system message

Sketch of the code change in `SummaryProvider.php`:

```php
// before
$systemPrompt = 'You are a helpful assistant that summarizes text in the same language as the text. '
              . 'You should only return the summary without any additional information.';

// after
$default = 'You are a helpful assistant that summarizes text in the same language as the text. '
         . 'You should only return the summary without any additional information.';
$systemPrompt = $this->appConfig->getAppValueString('summary_system_prompt', $default) ?: $default;
```

**Code Change 2 (related, same file):** consider also surfacing `max_tokens` for summarisation as a UI field. It's already tunable via `occ config:app:set integration_openai max_tokens --value=8000`, but the pairing (prompt + token budget) is what actually determines summary quality — surfacing both in the same UI section makes them discoverable together.

**Broader shape (nice-to-have, out of scope for this ticket):** the same "hardcoded prompt with no override" pattern likely applies to other `core:text2text:*` task types (headline, chat-summary, etc.). A broader "per-task-type prompt templates" epic would be the ideal end state, but even just the Summary field would unblock the most common request today.

### Describe alternatives you've considered

1. **Patching `SummaryProvider.php` directly on each server.** Works, but the patch is clobbered on every app upgrade. Not maintainable across a fleet, and every admin who wants a better summary has to re-derive the same patch.

2. **Using the existing "Chat User Instructions for Chat Completions" field.** That field only applies to the interactive chat path, not to TaskProcessing summarisation. Confirmed by inspecting `SummaryProvider::process()` — the chat instructions config is not consulted there.

3. **Using a middleware proxy (LiteLLM, LocalAI itself, etc.) to rewrite the system prompt.** LiteLLM can inject or replace system prompts on a per-model basis. This works, but pushes app-level configuration into infrastructure, is invisible to Nextcloud admins, and means every deployment needs a proxy in front of the LLM endpoint just for this one setting.

4. **Waiting for a broader "per-task-type prompt templates" feature.** Ideal end state, but shipping just the Summary override in the meantime would unblock most of the actual pain today.

Abhijeet-035:
## Summary

- add an admin-configurable summary system prompt
- preserve the existing default prompt when the setting is empty
- use the configured prompt in `SummaryProvider` while retaining format and complexity instructions
- add unit test coverage for the custom summary system prompt

Closes #405


lukasdotcom:
Hi @Abhijeet-035,

Thanks for the pr. There was a large pr recently, #431, Could you make sure to remove all the unrelated changes that probably came from that.

lukasdotcom:
Hi @Abhijeet-035,

You seem to have conflicts now.

lukasdotcom (approved):
I've fixed it up for you.

PR #435, opened 2026-09-11
fix: use max tokens parameter everywhere

lukasdotcom:
For thinking models limiting the max tokens does not work as thinking tokens are counted breaking these providers.

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI (N/A)


edward-ly (approved):
Makes sense to me 👍

